PR #462, opened 2015-02-10
WINDUP-489: Add default file-type mappings for common file-types

mbriskar:


windupgithubbot1:
https://10.3.9.192//job/windup-pr-builder/486/
Test PASSed.

rules-xml/src/main/java/org/jboss/windup/rules/apps/xml/DiscoverXmlFilesRuleProvider.java
```diff
@@ -49,8 +49,28 @@
...
+                    .addRule(FileMapping.from(".*\\.bcfg$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.map$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.brg$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.brgt$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.ruleset$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.module$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.modulex$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.composite$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.requirements$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.brgt$").to(XmlFileModel.class))
```
jsight: It looks like the last three are duplicates. Could you remove those?

rules-xml/src/main/java/org/jboss/windup/rules/apps/xml/DiscoverXmlFilesRuleProvider.java
```diff
@@ -49,8 +49,25 @@
     public Configuration getConfiguration(GraphContext context)
     {
         return ConfigurationBuilder.begin()
-                    
                     .addRule(FileMapping.from(".*\\.xml$").to(XmlFileModel.class))
+                    .addRule(FileMapping.from(".*\\.tld$").to(XmlFileModel.class))
```
lincolnthree: Good! But this should be implemented as a file-mappings.windup.xml rule file that gets included in our distribution. Please let me know if you have questions about how to do that!

dist/src/main/resources/rules/file-mappings.windup.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0"?>
+<ruleset xmlns="http://windup.jboss.org/v1/xml" id="XmlFileMappings">
+    <rules>
+        <rule>
```
jsight: These "<rule>" elements are unnecessary.

mbriskar: Well, yes, but I am following the Java class structure, in which they are necessary. I would rather have only one option of these configuration rules. Otherwise it may lead to a mess.

jsight: I don't think that I understand. file-mapping is a Rule, so putting it inside of a Rule doesn't completely make sense? I'm actually a little surprised that it works. AFAIC, the only reason that it does is that the RuleHandler class is a little lax in its handling of subelements.

jsight: To give an example of what I mean, take a look at the rule report. Each rule from this file will be preceded by a rule that looks something like this:

```
    "addRule()
    withId("GeneratedID_XmlFileMappings_1")
```

This is because the Rule adds a blank rule before processing the file-mapping Rule element.

lincolnthree: I agree with Jess. Remove the <rule/> wrappers :) file-mapping *is* a rule :)

dist/src/main/resources/rules/file-mappings.windup.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<ruleset xmlns="http://windup.jboss.org/v1/xml" id="XmlFileMappings">
+    <rules>
+           <file-mapping from=".*\.tld$" to="XmlFileModel"/>
```
lincolnthree: Beautiful.

lincolnthree:
Merged.

PR #465, opened 2015-02-11
WINDUP-482: Throw an exception in the case of rulesets with duplicate id...

jsight:
https://issues.jboss.org/browse/WINDUP-482

### DIFF
--- a/config/impl/src/main/java/org/jboss/windup/config/loader/WindupRuleLoaderImpl.java
+++ b/config/impl/src/main/java/org/jboss/windup/config/loader/WindupRuleLoaderImpl.java
@@ -2,7 +2,9 @@ package org.jboss.windup.config.loader;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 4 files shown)


config/impl/src/main/java/org/jboss/windup/config/loader/WindupRuleLoaderImpl.java
```diff
@@ -78,6 +80,20 @@ private void printRulePhases(List<WindupRuleProvider> allProviders)
...
 
+    private void checkForDuplicates(List<WindupRuleProvider> providers)
+    {
+        // using a map so that we can easily pull out the previous value later (in the case of a duplicate)
+        Set<WindupRuleProvider> dupeSet = new HashSet<>(providers.size());
+        for (WindupRuleProvider provider : providers)
+        {
+            if (dupeSet.contains(provider))
+            {
+                throw new WindupException("Found two providers with the same id: " + provider.getID());
```
lincolnthree: I wonder if this should also list the duplicate instances / types?

jsight: Yeah, possibly. Although I expect that most of the time the class will be WindupRuleProviderBuilder.

lincolnthree: True.

rules-java-archives/addon/pom.xml
```diff
@@ -11,7 +11,7 @@
     </parent>
 
     <artifactId>rules-java-archives</artifactId>
-    <name>Windup Ruleset: Archive Identification - Addon</name>
```
lincolnthree: Thank you.

PR #480, opened 2015-02-24
Fixes an issue in which an archive can be set with two different project...

jsight:
... models. This can produce strange results that are difficult to debug. Also, made it so that only the input directory is scanned, so we don't end up with user rules directories being scanned during Windup execution.


windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/526/
Test PASSed.

jsight:
@mbriskar I believe this fixes the reporting issue that we were discussing earlier.

PR #509, opened 2015-03-24
Set default $MAX_MEMORY

lincolnthree:


windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/628/
Test PASSed.

dist/src/main/resources/bin/windup
```diff
@@ -165,6 +165,10 @@ if [ -r "$WINDUP_HOME/addons/" ] ; then
   ADDONS_DIR="--immutableAddonDir $WINDUP_HOME/addons/"
 fi
 
+if [ -z "$MAX_MEMORY" ] ; then
+  MAX_MEMORY="2048m"
+fi
+
```
OndraZizka: Is that enough? I see JVM taking over 3 GB total. MAX_MEMORY goes to heap I guess? So the total could be around those 3 GB?

Anyway, machines with 4 GB could have around 4 GB of swap, so I think we could safely set 3 GB heap. Risk of swapping other processes out of RAM is better than risking performance degradation due to GC running all the time and evt. hitting OoMEx. my2c.

jsight: I'm still -1 on this patch (for reasons that I have mentioned previously), but I definitely wouldn't recommend going larger. Swapping and GC'ed languages combine extremely poorly for various reasons, and there are still a lot of 4GB languages out there.

lincolnthree:
Agreed.

PR #528, opened 2015-04-14
Update project: s/from/fromVersion and s/to/toVersion in the rule-schema

sgilda:


windupgithubbot1:

PR #535, opened 2015-04-27
Decompile filtering

bradsdavis:
Filters an archive's entries based on the --packages parameter.


bradsdavis:
Even though there is the ShouldDecompileCriterion, that will only tell you whether the JAR includes a package that is relevant to decompile, and thus if the JAR needs to be decompiled.  However, there is the situation (like super JARs) where there are both relevant and non-relevant classes in a JAR.  This filter skips non-relevant Java classes from decompiling from the archive.

windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/692/
Test PASSed.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/63/
Test PASSed.

jsight:
Continued in https://github.com/windup/windup/pull/536

PR #550, opened 2015-05-06
WINDUP-583 -  Correctly guesses non-ambiguous vendors by package name…

bradsdavis:
…. Updated organizations to leverage VendorModels within the graph rather than the project.  Moved LinkModel and ClassificationModel into parent project.

Many files changed because of refactoring LinkModel and ClassificationModel from reporting to be included in the WindupGraph project, since these should be more general than just for reporting.


jsight:
I don't believe that these should be moved to "Graph". Also, we are using "Organization" rather than "Vendor", and a lot of this is duplicating the existing package mapping code (but with hardcoded values instead).

windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/721/
Test FAILed.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/92/
Test FAILed.

bradsdavis:
I agree; I think Organization would be better since some aren't vendors.  

I am not clear on how PackageNameMapping works.

bradsdavis:
Also, @jsight I am not sure whether the Organization should be tied to the Project or Archive.  Do you have any recommendations there?  Right now, we have one per project, and it is tied to the Maven configuration.  But the problem is, I would like to disconnect the information within the individual configurations with the actual organization, so we can cleanse and standardize the data.  Thoughts?

jsight:
@bradsdavis Right now we are not pulling this out for non-maven projects, but we really should be. I think that is what this PR could do? ProjectModel is the abstract concept that you are looking for here, IMO as this is intended to group together logical units of code in an abstract sense.

I'm not sure what you mean by "disconnect the information within the individual configurations"?

bradsdavis:
@jsight right now, there is a ProjectModel and a MavenProjectModel, which extends ProjectModel.  But that, to me, indicates that the information comes from the Maven POM.

The problem is, we may get data from Maven, some data from the Manifest, and some more data from the project structure, and then need to cleanse and munge that together to figure out the best way to present that to the user.  How that "in-between" data is captured in the graph, and how that data is then cleansed and ultimately represented in its final form in the graph for reporting is what i need clarification on.

bradsdavis:
Also, we want to avoid inconsistencies.  Here is an example:

Project A: 
Pom says vendor is: The Apache Group

Project B:
Manifest says vendor is: Apache Organization

Project C:
No POM or Manifest, but it contains only org.apache

In all of those cases, each project artifact should ultimately say that the Organization is Apache.

jsight:
@bradsdavis Oh, I understand. Perhaps "MavenProjectModel" is a misleading term then. I don't personally think that all of that data needs to have a 1:1 relationship with the POM file. If we need to model the POM directly in order to maintain the literal descriptive data for some reason, that should be separate from the ProjectModel itself, IMO (for example, something like MavenPomModel that itself is linked to from the Project).

Basically what I am saying is that if you feel like you can replace literal POM information with information that is functionally more accurate, you shouldn't be afraid to do so. Maybe MavenProjectModel isn't the best name for this model?

bradsdavis:
@jsight I think we will need some higher level rules.

For example, we should have the discovery rule for Manifests, and POM files that extract base information.  Then, we should have a rule set later in the rule set that takes the POM info, and tries to cleanse it for things like organization.  We should do the same with the manifest information.  Then, we pull the information in the end that is available.  

I do agree that we need to remove MavenProjectModel, and make it a MavenConfigurationModel, which extends XmlFileModel.  That would be better in my opinion.  Thoughts?

jsight:
Yes, additional rules to groom the data based upon all available information makes sense to me.

I don't agree with removing MavenProjectModel and never said that. :) MavenPomModel would be in addition to that. It's useful to be able to group a project together in an abstract sense (not just at the archive level).

bradsdavis:
@jsight Regarding moving LinkModel and ClassificationModel into the org.jboss.windup.graph.model package, I don't see those as any different than ProjectModel.  Can you clarify why one should be in org.jboss.windup.graph.model and the other in org.jboss.windup.reporting.service?

LinkModel can be used for reporting, but it may also be used to add a link to the Organizations HQ URL, which isn't for reporting, that is just informational.

bradsdavis:
Currently, we have to also put these cleansing rules in Java, but really should be extracting these kinds of things out to be XML driven in the future.

jsight:
@bradsdavis The location of ProjectModel is also wrong. rules-base would probably be a better place for them. I'm still not sure what LinkModel is going to be used for other than reporting to the user.

jsight:
@bradsdavis Vendor mapping is configured in the XML:

https://github.com/windup/windup-rulesets/blob/master/rules/filemappings/default-package-to-organization-names.windup.xml

bradsdavis:
How can you have linkmodel off of organizationmodel or classificationmodel if those were in graph and the others were in reports projects? There is no dependency visibility between the two.  Link can be used for reporting, but not specifically only html reporting. Eclipse tooling might want to know about an organizations link for the user experience for example.

jsight:
Eclipse tooling would not be using either of these interfaces directly. As I said earlier, I think ProjectModel is in the wrong place. It should probably be in rules-base. I could see where linkmodel could go into there as well, but probably not classificationmodel.

windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/789/
Test FAILed.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/156/
Test FAILed.

windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/790/
Test FAILed.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/157/
Test FAILed.

windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/791/
Test FAILed.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/158/
Test FAILed.

lincolnthree:
@bradsdavis I tried to merge this, but it looks like there are files missing. (The stuff that actually does the work.) Can you check for those commits in your git reflog and try to add them back in? (Or just paste them in again from history if that is simpler.)

lincolnthree:
Merged.

PR #561, opened 2015-05-21
Change rule-schema.xsd XML version to 1.0

sgilda:


windupgithubbot1:

windupgithubbot1:

lincolnthree:
Why is this necessary?

jsight:
I think that it renders correctly in Firefox after this change, and I don't know of any downsides.

If noone objects, I am going to go ahead and merge.

sgilda:
+1

mareknovotny:
just to sum up this discussion, XML version 1.1 is not backward compatible with version 1.0. Some parsers could have issue - one simple use case but with no big crucial effect of that is the Firefox browser can't render the XML rule file.

So I read if there is no direct requirement to use 1.1 feature(s) it is recommended to use version 1.0. 
http://www.w3.org/TR/xml11/#sec-xml11

> The overall philosophy of names has changed since XML 1.0. Whereas XML 1.0 provided a rigid definition of names, wherein everything that was not permitted was forbidden, XML 1.1 names are designed so that everything that is not forbidden (for a specific reason) is permitted. Since Unicode will continue to grow past version 4.0, further changes to XML can be avoided by allowing almost any character, including those not yet assigned, in names.

Anyway I can understand we could need this version.

PR #582, opened 2015-06-01
Rename XSD, put it under /schema, package it to .jar. For WINDUP-612 …

OndraZizka:
… Validate all XML rules in windup-rulesets during build.

Hopefully done correctly this time.
Needed by https://github.com/windup/windup-rulesets/pull/17 .
This assumes that the addon will contain all schemas released so far, hence the location in schema/


config-xml/addon/pom.xml
```diff
@@ -50,6 +50,17 @@
     </dependencies>
 
     <build>
+        <resources>
+            <resource>
+                <directory>../schema/</directory>
+                <includes>
+                    <include>windup-ruleset-jboss-${project.version}.xsd</include>
```
lincolnthree: rename to windup-jboss-ruleset.xsd

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/171/
Test PASSed.

config-xml/addon/pom.xml
OndraZizka: I am keeping non-versioned, a copy just for the sake of the version being in the file name is not needed, at least not now.

Later, we could perhaps externalize the schema version to root POM and use ${...} inside of the schema, with resources filtering; then, copying + filtering may be useful.

config-xml/addon/pom.xml
```diff
@@ -50,8 +50,36 @@
...
+            </resource>
+        </resources>
         <plugins>
             <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>1.8</version>
+                <executions>
+                    <execution>
+                        <phase>generate-resources</phase>
+                        <goals> <goal>run</goal> </goals>
```
lincolnthree: This formatting is a bit odd. Shouldn't there be newlines here?

PR #629, opened 2015-07-03
Test script to build and run the latest windup

mbriskar:
Inspired after reviewing PR I had an idea of creating a common directory for the scripts that we use for testing purposes. They may be then placed in some test{?} directory along with test-files  directory.

The script that I have build builds the current windup and then executes it with the rulesets and distribution of the same version (hopefully) and run the result on javaee application. This is mainly useful **when changing the layout**, after having it done. all you need to do is to care about the code and then run the script and get the output.

### DIFF
--- a/test-scripts/build_all_and_run_javaee.sh
+++ b/test-scripts/build_all_and_run_javaee.sh
@@ -1,0 +1,39 @@
+#run this if you changed only the windup/windup and not distribution nor ruleset. Then it will load the appropriate versions of ruleset/distribution and run them on javaee example application
+
+cd ..
+#run the current branch
+mvn clean install -DskipTests
+#get current version
+version=`mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version|grep -Ev '(^\[|Download\w+:)'`
+
+#build rulesets
+cd ../windup-rulesets
+git fetch --all
+git rebase upstream/master
+rulesetVersion=`mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version|grep -Ev '(^\[|Download\w+:)'`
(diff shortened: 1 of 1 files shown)


windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/910/
Test PASSed.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/279/
Test PASSed.

lincolnthree:
While I like the idea of keeping a directory of scripts. I'm not sure the file you've included in this PR is appropriate for checkin. It assumes a folder and path structure that you've laid out that other members of the team might not.

mbriskar:
I understand, however I don't see a way how to overcome this

jsight:
I'm not sure either. It seems like it would be difficult to make this environment agnostic without having it check everything out on its own.

pmuir:
Use default values in the script that can be overridden with env vars? in bash use ={VAR_NAME:default_value}

pmuir:
What about using default values that can be overridden using bash env vars. This is done using

FOO=${FOO:default_value}

mbriskar:
another possible approach is to use grep .. --include *pom.xml and find the place where are these poms. It would not work if there are multiple projects. 

So I like the default properties

jsight:
I'm actually more bothered by the fact that it does fetch and rebase operations on other repos. I'm not sure if that is a good default behavior?

mbriskar:
I used this script to check the PR that changed the layout. But it is true. I may probably save a branch into variable, checkout upstream/master and then checkout back

OndraZizka:
Hi, check the windup-ci repository which I created for this purpose.

mbriskar:
Thanks, I am moving this in the windup-ci. I will download everything that is not needed in this script so it will not interfere with other repos.

PR #662, opened 2015-08-25
Update the message in RulesetUpdateChecker.java.

OndraZizka:


ui/addon/src/main/java/org/jboss/windup/ui/RulesetUpdateChecker.java
```diff
@@ -32,7 +32,7 @@ public void perform(@Observes PostStartup event)
         if (updater.rulesetsNeedUpdate())
         {
             System.out.println("\nThe rulesets are outdated: " + updater.getRulesetsDir()
-                + "\nConsider running Windup with --updateRuleset or --updateDistribution. Press ENTER to continue.\n");
+                + "\nConsider running Windup with --updateRuleset or --updateDistribution.\n");
```
jsight: Hmmm, won't this appear even in cases where the user already specified "--updateRuleset"? Also, do we even support --updateDistribution?

PR #719, opened 2015-10-07
WINDUP-798: Ensure that IP addresses are identifed for only n.n.n.n p…

jsight:
https://issues.jboss.org/browse/WINDUP-798

### DIFF
--- a/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverStaticIPAddressRuleProvider.java
+++ b/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverStaticIPAddressRuleProvider.java
@@ -66,7 +66,7 @@ public class DiscoverStaticIPAddressRuleProvider extends AbstractRuleProvider
                             }
                         }
                     })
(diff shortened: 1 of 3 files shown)


rdruss:
Should a test be added to ensure an n.n.n.n.n pattern is *not* matched?

rdruss:
Would the unexpected match be this:

unexpectedIPs.add("192.168.0.9.3.4");
-or-
unexpectedIPs.add("192.168.0.9");

I know in the false positives in the report, it looks like a valid IP address until you look at the source.

jsight:
In theory, it could be either one or even something else entirely. Having said that, there is an assertion at the end to try to catch this:

Assert.assertEquals(expectedIPs.size(), numberFound);

Basically, if we get anything that is in the expected list we fail, and we also fail if we get anything that is in neither list.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/471/
Test PASSed.

windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/1102/
Test PASSed.

windupgithubbot1:
http://10.3.8.5//job/windup-core-PR-dependents-check/472/
Test PASSed.

windupgithubbot1:
http://10.3.8.5//job/windup-pr-builder/1103/
Test PASSed.

PR #782, opened 2015-12-07
WINDUP-863 - Made the static IP Address scanner take into considerati…

bradsdavis:
…on commented out properties and Maven version numbers.

### DIFF
--- a/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverStaticIPAddressRuleProvider.java
+++ b/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverStaticIPAddressRuleProvider.java
@@ -1,5 +1,15 @@
 package org.jboss.windup.rules.apps.java.ip;
 
+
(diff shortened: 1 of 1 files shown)


windupgithubbot1:
http://wonka.mw.lab.eng.bos.redhat.com/jenkins//job/windup-pr-builder/1269/

windupgithubbot1:
http://wonka.mw.lab.eng.bos.redhat.com/jenkins//job/windup-core-PR-dependents-check/643/

bradsdavis:
@jsight When this was merged, the isMavenVersionTag method was removed and it is dead code right now.  However, I think it should still be included in the code.  

Specifically, you may have this XML: 

```
<version>
1.3.5.6
</version>
```

That would not be caught in the ignoreLine method alone because version is not on the same line as the 1.3.5.6 value -- since it is XML.  I think this is very likely going to be encountered in Maven files, which is why I handle Maven differently than property files.

jsight:
@bradsdavis - Good point... I'll merge a fix. Thanks for the contribution!

PR #785, opened 2015-12-07
WINDUP-867 - Updated to keep memory of the expansion of the overview …

bradsdavis:
…tabs


windupgithubbot1:
http://wonka.mw.lab.eng.bos.redhat.com/jenkins//job/windup-core-PR-dependents-check/646/

windupgithubbot1:
http://wonka.mw.lab.eng.bos.redhat.com/jenkins//job/windup-pr-builder/1272/

jsight:
How is this scoped? Won't sessage storage essentially have the side effect of leaking to other reports (potentially from other applications)? For example, if I look at application A and open a few panels, then look at application B it's possible for the ids to overlap?

bradsdavis:
It's session scoped.

bradsdavis:
Session Storage:

Values persist only as long as the window or tab in which they stored.

Values are only visible within the window or tab that created them.

mareknovotny:
@bradsdavis I see that the change broke the initial rendering of report. It should be all collapsed, but it is rendered now as all Expanded and the link for switching of expand/collapse all is not reflecting the current status - it shows Expand All while all is expanded. 
And the functionality of saving the position for back navigation is due that expanded all anyway broken :(

jsight:
I would also suggest using a GUID instead of the vertex id. It would be less likely to have collisions across multiple reports.

bradsdavis:
@jess I don't think it will affect across reports since it is session scoped. 

@marek it shouldn't all be collapsed when you first load the report. It should only collapse the 0 story point jars. The others I'm interested in reviewing and meet to be open so I don't need to search through them. In a large app, finding non 0 jars of interest is difficult when they are all collapsed, and also difficult when they are all expanded. Only expanding non 0 story points by default is the best solution and we have a button to expand / collapse all at once if someone is interested in collapsing them.

jsight:
@bradsdavis "Session scoped" in this context is just within a single tab in the browser. If they reuse that tab for multiple reports, then the data is shared.

jsight:
@rdruss What is your opinion regarding the defaults for open vs closed?

bradsdavis:
I can share a large customer app as an example if you guys need one. That's what influenced my opinion.

mareknovotny:
@bradsdavis  if that should collapse only panels with 0 SP then it doesn't work on my build. What browsers have you tested?

jsight:
@bradsdavis - I'm open to the idea, I just want to make sure that we get feedback from the rest of the team. I can see it being useful on some of the larger apps that we test with.

mareknovotny:
btw, @mbriskar added the feature of collapsing to get easier overview on the report, so even @bradsdavis says it should collapse only non-interesting panels with 0 SP, it still will produce long report by default, I don't argue against caching of already reviewed report and going back to it, just default first visit/render of the report.

rdruss:
I like the idea that @bradsdavis has proposed.  I think the 0 story point jar sections are the noise that we're trying to eliminate.  When I've experimented with the feature, I have found myself looking at the overall report and expanding anything that had greater than 0 story points.

jsight:
Merged -- seems to work fine here. I did tweak it to use the guid as otherwise it would leak across pages (even leaking across the overview and catchall page).

mareknovotny:
ok, I still see broken link "Expand/Collapse All" as it doesn't take into count or don't check if all panels are expanded it still supply Expand All instead of Collapse All link :(

Be more clear - if some panels are collapsed the behaviour is correct, but if they are all expanded the link stays as Expand All instead of Collapse All.

bradsdavis:
Can you submit a patch for the expand all / collapse all issue you are seeing?

mareknovotny:
I will try to fix it. noting myself - I saw that behaviour on org.jboss.windup.tests.application.WindupArchitectureCatchallTest there is only one

mareknovotny:
I sent PR https://github.com/windup/windup/pull/792 this should fix it

PR #818, opened 2016-01-27
fix InstallRemoveAddonCommandsTest so that it doesn't mess up with ~/.windup/.addons

Ladicek:


jsight:
Thanks @Ladicek!

Ladicek:
Sorry for causing the mess. This should fix it, though `~/.windup/.addons` on the CI server needs to be cleaned.

jsight:
No worries, the tests themselves are great (and something that we have really needed for a long time).

PR #829, opened 2016-02-01
WINDUP-853: The --discoverPackages output seems wrong (fixed the sort…

jsight:
https://issues.jboss.org/browse/WINDUP-853


mareknovotny:
sorting is correct for Known packages.

PR #842, opened 2016-02-15
Cosmetic changes, mostly log-messages related.

OndraZizka:


bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/RunWindupCommand.java
```diff
@@ -228,29 +233,13 @@ else if (Boolean.class.isAssignableFrom(option.getType()))
...
-            System.out.println("Input Applications:");
-            for (Path inputPath : windupConfiguration.getInputPaths())
-            {
-                System.out.println("\t" + inputPath);
-            }
-            System.out.println();
-        }
-        System.out.println("Output Path:" + windupConfiguration.getOutputDirectory());
-        System.out.println();
-
```
OndraZizka: Should this be pronted in RunWindupCommand? I think it should rather be printed in WindupProcessorImpl so it's printed in every execution path.

jsight: This goes to the user when running from the command line. WindupProcessorImp will just end up in a log file.

OndraZizka: Does the user need to see this few lines below the input command?

OndraZizka: I have left it there, no need to block this PR with that debate :)

OndraZizka:
retest

OndraZizka:
Also I've changed the title - this actually had not much to do with WINDUP-885.

PR #843, opened 2016-02-15
    WINDUP-922  Make the package prefixes ignore a .* at the end

OndraZizka:


jsight:
If we are going to do this, it would be better to do it in the UI and print a warning to the user. It isn't a package pattern.

OndraZizka:
That's right, but at which UI? The common execution path starts in WindupProcessorImpl.

jsight:
WindupProcessorImpl isn't UI at all. The UI should have validation for this kind of thing and how data gets into windup execution will vary based upon the type of UI.

Some will likely have graphical interfaces for package selection.

OndraZizka:
I have moved it to RunWindupCommand.
For the Maven plugin, it's in https://github.com/windup/windup-maven-plugin/pull/5

PR #854, opened 2016-02-19
[WINDUP-937] fix the link to additional resources in Migration Issues report

Ladicek:


reporting/impl/src/main/resources/reports/templates/migration-issues.ftl
```diff
@@ -328,7 +328,7 @@
                                         <div class="panel-body">
                                             <ul>
                                                 {{#each ../resourceLinks}}
-                                                    <li><a href="{{link}}">{{title}}</a></li>
+                                                    <li><a href="{{href}}">{{title}}</a></li>
```
Ladicek: See line 409 of this file which constructs the relevant portion of the JSON file.

jsight:
Thanks @Ladicek!

PR #858, opened 2016-03-16
Tweaked to also mark 'implements' for interfaces implemented by superclasses

jsight:
@mareknovotny - This doesn't completely solve your problem, but maybe it can be a starting point?

### DIFF
--- a/java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
+++ b/java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
@@ -13,7 +13,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.eclipse.jdt.core.dom.ASTVisitor;
 import org.eclipse.jdt.core.dom.AbstractTypeDeclaration;
 import org.eclipse.jdt.core.dom.Annotation;
-import org.eclipse.jdt.core.dom.AnnotationTypeDeclaration;
 import org.eclipse.jdt.core.dom.AnonymousClassDeclaration;
 import org.eclipse.jdt.core.dom.ArrayInitializer;
 import org.eclipse.jdt.core.dom.BooleanLiteral;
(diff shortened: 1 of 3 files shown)


mareknovotny:
ok. I tried that without success. 
What I am now thinking about it is the requirement to add EAP7 metadatas as we have for EAP (6) api jars and it seems like a boiler plate to adding all api jars for eap 7 in the same way we have for eap6.
We probably need to implement some metadata index resolvers not to rely on existing jars/classes on classpath.

WDYT? @jsight @OndraZizka

jsight:
@mareknovotny - I don't know what you mean about trying it without success. I was able to use that rule yesterday, I just had to have it target eap (and I had to hardcode the test harness to target eap as well).

I agree that we need appropriate jars for EAP7. I'm not sure what you mean about metadata resolvers. Can JDT bind classnames based upon something other than API jars being available?

mareknovotny:
@jsight I meant that my test case for the windup rule failed even I add targetTechnology to the ruleset where the rule was with this PR. I didn't know I need to change target for my test rule too. Test rules don't have metadata with target(s), where did you add that eap target or is that just the same as in real windup rules?

> I agree that we need appropriate jars for EAP7. I'm not sure what you mean about metadata resolvers. Can JDT bind classnames based upon something other than API jars being available?

First thing is to detached api-jars from windup-ruleset project (to be scalable in adding more APi target versions and reduce that artifact size) and do some assembly of binaries only as dependencies in windup distribution IMHO.
Next and probably more appropriate thing is to not rely on classpath jars when you are resolving class hierarchy, but create an index with class relations or hierarchy without real java classes. As @OndraZizka commented somewhere we don't need method/classes bodies, we need only their declarations to resolve/achieve/search the objects hierarchy.

jsight:
@mareknovotny - I don't mean in the rules themselves. The test runner itself runs with no target specified. It had to be modified to hardcode the target of the test runner as "eap" as well.

I agree that API jars should not have implementations. I disagree that this means the addition of another file format, though. I think you would basically find out that you are reimplementing .class files as well as a mechanism for loading .class file information into the Eclipse compiler if you go down that road.

Better would be to build a tool to transform .class files with implementations into smaller stubs with only the API surface visible. Existing Java tools (including JDT) are used to dealing with .class files. They aren't used to dealing with whatever index format we would come up with.

mareknovotny:
@jsight ok, I got.

+1 for a tool to cut the implementation bodies from jars.

jsight:
@mareknovotny - I have attached a small tweak to change the handling of TYPE declarations that are recovered (as opposed to resolved). I'm not sure if this will help with the issues that you were having?

mareknovotny:
so I haven't found any issue with this change, so I guess we can merge it.

jsight:
@mareknovotny - Ok, I'll go ahead and merge. I suspect that we still need more changes to support annotation queries properly.

PR #860, opened 2016-03-21
WINDUP-952 Multi-application support for Spring bean report

mkouba:
Note that this PR does not contain a test. Actually, I'm not quite sure how to write such a test yet. But I will try to do my best if it's required ;-)

### DIFF
--- a/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/service/SpringBeanService.java
+++ b/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/service/SpringBeanService.java
@@ -1,14 +1,19 @@
 package org.jboss.windup.rules.apps.javaee.service;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 import org.jboss.windup.graph.GraphContext;
+import org.jboss.windup.graph.model.ProjectModel;
 import org.jboss.windup.graph.service.GraphService;
 import org.jboss.windup.rules.apps.javaee.model.SpringBeanModel;
 
 /**
  * Contains methods for creating, querying, and updating SpringBeanModel entries in the Graph.
- * 
+ *
  * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
- * 
+ *
  */
 public class SpringBeanService extends GraphService<SpringBeanModel>
 {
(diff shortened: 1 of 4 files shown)


windupgithubbot1:

windupgithubbot1:

jsight:
@windupgithubbot1 retest

jsight:
@windupgithubbot1 retest

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/service/SpringBeanService.java
```diff
@@ -21,4 +26,22 @@ public SpringBeanService(GraphContext context)
...
+     */
+    public List<SpringBeanModel> findAllByApplication(ProjectModel application)
+    {
+        List<SpringBeanModel> models = new ArrayList<>();
+        for (SpringBeanModel model : findAll())
+        {
+            if (application.equals(model.getApplication()))
+            {
+                models.add(model);
+            }
```
OndraZizka: It would be better to filter at the DB query level. Imagine getting 1 000 000 spring beans from 1000 apps. That means 1000 x running through 1 000 000 beans filtering out a fraction of them. Not that I expect 1mil beans, but we should pay attention to performance.

jsight: @OndraZizka +1 - I changed this while merging.

mkouba:
Makes sense. I was not sure how to implement this properly. Anyway, we should probably review all other rules as well. For instance, `CreateJPAReportRuleProvider` gets and filters all the JPA entities...

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/service/SpringBeanService.java
mkouba: Makes sense. I was not sure how to implement this properly. Anyway, we should probably review all other rules as well. For instance, CreateJPAReportRuleProvider gets and filters all the JPA entities...

jsight: Yes, those could likely be improved as well. I did go through them to find any cases where we weren't filtering by app (hopefully I found them all).

PR #861, opened 2016-03-21
WINDUP-953 Generate Jandex indices to speed up Furnace bootstrap

mkouba:


windupgithubbot1:

windupgithubbot1:

jsight:
Is there any indication of how much of a difference this made?

mkouba:
No exact numbers. And I would guess it won't be that significant. But AFAIK this it what Forge guys did to improve the bootstrap and it made Weld classes disappear from the list of hot spots.

jsight:
@mkouba I tested it a few times and could not see a single second of improvement when running against jee-example.

It makes me wonder if it is actually working?

mkouba:
It should work according to the log. I will try to do some tests...

mkouba:
I cannot see any improvement neither. Actually, Weld SE is using Jandex even if there is no index available (it's created on the fly). Also Jandex only helps to filter out classes which should not end up as managed beans and to filter `ProcessAnnotatedType` observers with `@WithAnnotations` specified (and unfortunately neither is the case of Forge/Windup).

I'm going to close this pull request and corresponding JIRA as we're not observing any improment.

PR #866, opened 2016-04-05
Fix labels - Severity, Effort

OndraZizka:


jsight:
I don't think that this was what was requested.

PR #869, opened 2016-04-11
Added rule override functionality

jsight:


config/api/src/main/java/org/jboss/windup/config/RuleSubset.java
```diff
@@ -190,7 +190,7 @@ public void perform(Rewrite rewrite, EvaluationContext context)
             throw new IllegalArgumentException("Rewrite must be an instanceof GraphRewrite");
 
         /*
-         * Highly optimized loop - for performance reasons. Think before you change this!
+         * Highly optimized loop - for performance reasons. Think before you change this! (lincolnthree)
          */
```
OndraZizka: It's a highly optimized loop that we only go once through for each rule :)
This is a copy & paste from Rewrite IIRC, and should be refactored somehow so that we don't need to have a clone but rather an override.

config/impl/src/main/java/org/jboss/windup/config/loader/RuleLoaderImpl.java
```diff
@@ -131,12 +132,24 @@ private void checkForDuplicateProviders(List<RuleProvider> providers)
 
     private RuleProviderRegistry build(GraphContext context, Predicate<RuleProvider> ruleProviderFilter)
     {
-
-        ConfigurationBuilder result = ConfigurationBuilder.begin();
+        List<Rule> allRules = new ArrayList<>(2000); // estimate of how many rules we will likely see
```
OndraZizka: Can you please describe this change?

config/api/src/main/java/org/jboss/windup/config/RuleSubset.java
jsight: Yes, basically. It's optimization isn't so important for us, IMO.

It is not an exact clone, though, so I think it makes more sense here than as an override.

config/impl/src/main/java/org/jboss/windup/config/loader/RuleLoaderImpl.java
```diff
@@ -147,8 +160,26 @@ private RuleProviderRegistry build(GraphContext context, Predicate<RuleProvider>
...
+            // copy it to allow for the option of modification
+            List<Rule> rules = new ArrayList<>(cfg.getRules());
+            ListIterator<Rule> ruleIterator = rules.listIterator();
+            while (ruleIterator.hasNext())
+            {
+                Rule rule = ruleIterator.next();
+                Rule overrideRule = overrideRules.get(new RuleKey(provider.getMetadata().getID(), rule.getId()));
+                if (overrideRule != null)
+                {
+                    LOG.info("Replacing rule: " + rule.getId() + "with a user override!");
```
OndraZizka: Missing a space before "with "

config/impl/src/main/java/org/jboss/windup/config/loader/RuleLoaderImpl.java
jsight: It is just separating the creation of the Configuration from the rule list to allow the rule list to be configured independently.

config/impl/src/main/java/org/jboss/windup/config/loader/RuleLoaderImpl.java
jsight: Good catch.

config/tests/src/test/java/org/jboss/windup/config/RuleProviderOverrideTest.java
```diff
@@ -0,0 +1,144 @@
...
+                            return "OriginalRule";
+                        }
+                    });
+        }
+
+        @Singleton
+        public static class TestOverrideProvider extends AbstractRuleProvider {
+            public TestOverrideProvider() {
+                super(MetadataBuilder.forProvider(TestOverrideProvider.class, "TestRuleProvider").setOverrideProvider(true));
+            }
```
OndraZizka: This will collide with my PR (just a note)

jsight: That is a new test. I'm not sure why that it would?

OndraZizka:
Jess,
1) Could you perhaps add a test case with a XML file ruleset being overriden?
2) Where's the documentation

config-xml/addon/src/main/java/org/jboss/windup/config/parser/metadata/MetadataProviderOverrideHandler.java
```diff
@@ -0,0 +1,34 @@
...
+ *     &lt;overrideProvider&gt;
+ *          true
+ *     &lt;/overrideProvider&gt;
+ * </pre>
+ */
+@NamespaceElementHandler(elementName = MetadataProviderOverrideHandler.OVERRIDE_PROVIDER, namespace = RuleProviderHandler.WINDUP_RULE_NAMESPACE)
+public class MetadataProviderOverrideHandler implements ElementHandler<Boolean>
+{
+    public static final String OVERRIDE_PROVIDER = "overrideProvider";
+
```
OndraZizka: How about "<overridingRuleset>"?  Provider is a term that we didn't promote in the documentation, AFAIK, or is it?

OndraZizka:
WRT docs - how about "Indicates that this ruleset will override any rules distributed with Windup with matching IDs"?

jsight:
@OndraZizka - A test case for XML has been added

config/tests/src/test/java/org/jboss/windup/config/RuleProviderOverrideTest.java
OndraZizka: Because the @RuleMetadata processing overrides the metadata, IIRC. Not sure, would have to check.

OndraZizka: Actually it's ok. Just checked. The forProvider() itself reads the annotation and then the customizations happen.

PR #878, opened 2016-04-20
61 Mavenize (MIGR-233)

OndraZizka:
See taiga for what's left to consider it done.
However it already gives some usable results.

### DIFF
--- a/config/api/src/main/java/org/jboss/windup/config/query/QueryTypeCriterion.java
+++ b/config/api/src/main/java/org/jboss/windup/config/query/QueryTypeCriterion.java
@@ -8,8 +8,8 @@ import org.jboss.windup.graph.model.WindupVertexFrame;
 import com.tinkerpop.blueprints.Predicate;
 import com.tinkerpop.blueprints.Vertex;
 import com.tinkerpop.frames.FramedGraphQuery;
-import com.tinkerpop.frames.modules.typedgraph.TypeValue;
 import com.tinkerpop.gremlin.java.GremlinPipeline;
+import org.jboss.windup.graph.frames.TypeAwareFramedGraphQuery;
 
 class QueryTypeCriterion implements QueryFramesCriterion, QueryGremlinCriterion
 {
@@ -19,7 +19,7 @@ class QueryTypeCriterion implements QueryFramesCriterion, QueryGremlinCriterion
     public QueryTypeCriterion(Class<? extends WindupVertexFrame> clazz)
     {
         this.searchedClass = clazz;
-        this.typeValue = getTypeValue(clazz);
+        this.typeValue = TypeAwareFramedGraphQuery.getTypeValue(clazz);
     }
 
     @Override
@@ -28,18 +28,6 @@ class QueryTypeCriterion implements QueryFramesCriterion, QueryGremlinCriterion
         q.has(WindupVertexFrame.TYPE_PROP, typeValue);
     }
 
-    private static String getTypeValue(Class<? extends WindupVertexFrame> clazz)
-    {
-        TypeValue typeValueAnnotation = clazz.getAnnotation(TypeValue.class);
-        if (typeValueAnnotation == null)
-        {
-            throw new IllegalArgumentException("Class " + clazz.getCanonicalName() + " lacks a @TypeValue annotation");
-        }
-        else
-        {
-            return typeValueAnnotation.value();
-        }
-    }
 
     /**
      * Adds a criterion to given pipeline which filters out vertices representing given WindupVertexFrame.
@@ -47,7 +35,7 @@ class QueryTypeCriterion implements QueryFramesCriterion, QueryGremlinCriterion
     public static GremlinPipeline<Vertex, Vertex> addPipeFor(GremlinPipeline<Vertex, Vertex> pipeline,
                 Class<? extends WindupVertexFrame> clazz)
     {
-        pipeline.has(WindupVertexFrame.TYPE_PROP, getTypeValue(clazz));
+        pipeline.has(WindupVertexFrame.TYPE_PROP, TypeAwareFramedGraphQuery.getTypeValue(clazz));
         return pipeline;
     }
 
--- a/decompiler/impl-fernflower/src/main/java/org/jboss/windup/decompiler/fernflower/FernflowerDecompiler.java
+++ b/decompiler/impl-fernflower/src/main/java/org/jboss/windup/decompiler/fernflower/FernflowerDecompiler.java
@@ -11,9 +11,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
@@ -25,12 +22,8 @@ import org.jboss.windup.decompiler.api.DecompilationException;
 import org.jboss.windup.decompiler.api.DecompilationFailure;
 import org.jboss.windup.decompiler.api.DecompilationListener;
 import org.jboss.windup.decompiler.api.DecompilationResult;
-import org.jboss.windup.decompiler.api.Decompiler;
 import org.jboss.windup.decompiler.decompiler.AbstractDecompiler;
 import org.jboss.windup.decompiler.util.Filter;
-import org.jboss.windup.util.Checks;
-import org.jboss.windup.util.threading.WindupChildThreadFactory;
-import org.jboss.windup.util.threading.WindupExecutors;
 import org.jetbrains.java.decompiler.main.Fernflower;
 import org.jetbrains.java.decompiler.main.extern.IBytecodeProvider;
 import org.jetbrains.java.decompiler.main.extern.IFernflowerPreferences;
@@ -165,7 +158,8 @@ public class FernflowerDecompiler extends AbstractDecompiler
 
 
 
-    private List<String> pathsFromDecompilationRequests(List<ClassDecompileRequest> requests) {
+    private List<String> pathsFromDecompilationRequests(List<ClassDecompileRequest> requests)
+    {
         List<String> result = new ArrayList<>();
         for(ClassDecompileRequest request : requests) {
             result.add(request.getClassFile().toString());
--- a/graph/api/src/main/java/org/jboss/windup/graph/GraphContext.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/GraphContext.java
@@ -14,7 +14,7 @@ import org.jboss.windup.graph.service.GraphService;
 
 /**
  * Context for interacting with the underlying graph database API.
- * 
+ *
  * @author <a href="mailto:lincolnbaxter@gmail.com">Lincoln Baxter, III</a>
  * @author <a href="mailto:zizka@seznam.cz">Ondrej Zizka, I</a>
  */
@@ -29,18 +29,18 @@ public interface GraphContext extends Closeable
      * Get the underlying {@link EventGraph}, which is itself a wrapper for a {@link TitanGraph}.
      */
     EventGraph<TitanGraph> getGraph();
-    
+
     /**
      * Creates new graph using the configuration. In case there was already a graph located in the specified path, it will be deleted.
      */
     GraphContext create();
-    
+
     /**
      * Loads the graph using the configuration.
      */
     GraphContext load();
-    
-    
+
+
 
     /**
      * Get the {@link FramedGraph} view of the underlying {@link EventGraph}.
@@ -68,7 +68,12 @@ public interface GraphContext extends Closeable
     void setOptions(Map<String, Object> options);
 
     /**
-     * Returns the globally configured options as an immutable {@link Map}
+     * Returns the globally configured options as an immutable {@link Map}.
+     *
+     * Example usage:
+     * <pre>
+     * Boolean overwrite = (Boolean) windupConfiguration.getOptionMap().get(OverwriteOption.NAME);
+     * </pre>
      */
     Map<String, Object> getOptionMap();
 
--- a/reporting/api/src/main/java/org/jboss/windup/reporting/freemarker/FurnaceFreeMarkerTemplateLoader.java
+++ b/reporting/api/src/main/java/org/jboss/windup/reporting/freemarker/FurnaceFreeMarkerTemplateLoader.java
@@ -11,7 +11,7 @@ import org.jboss.windup.config.furnace.FurnaceHolder;
 
 import freemarker.cache.URLTemplateLoader;
 
-class FurnaceFreeMarkerTemplateLoader extends URLTemplateLoader
+public class FurnaceFreeMarkerTemplateLoader extends URLTemplateLoader
 {
     @Override
     protected URL getURL(final String name)
--- a/rules-java-archives/addon/pom.xml
+++ b/rules-java-archives/addon/pom.xml
@@ -15,11 +15,30 @@
 
         <!-- Local Dependencies -->
         <dependency>
+            <groupId>org.jboss.windup.maven</groupId>
+            <artifactId>nexus-indexer-data</artifactId>
+            <version>6-SNAPSHOT</version>
+            <scope>compile</scope>
+            <optional>true</optional>
+            <exclusions>
+                <exclusion>
+                    <groupId>commons-codec</groupId>
+                    <artifactId>commons-codec</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.lucene</groupId>
+                    <artifactId>lucene-core</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <!-- The API from nexus-indexer-data could make this unneeded. -->
+        <dependency>
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-core</artifactId>
             <version>4.8.1</version>
             <scope>provided</scope>
         </dependency>
+
 
         <!-- Addon Dependencies -->
         <dependency>
--- a/rules-java-archives/addon/pom.xml
+++ b/rules-java-archives/addon/pom.xml
@@ -17,7 +17,7 @@
         <dependency>
             <groupId>org.jboss.windup.maven</groupId>
             <artifactId>nexus-indexer-data</artifactId>
-            <version>6-SNAPSHOT</version>
+            <version>6</version>
             <scope>compile</scope>
             <optional>true</optional>
             <exclusions>
--- a/rules-java-archives/addon/pom.xml
+++ b/rules-java-archives/addon/pom.xml
@@ -17,7 +17,7 @@
         <dependency>
             <groupId>org.jboss.windup.maven</groupId>
             <artifactId>nexus-indexer-data</artifactId>
-            <version>6</version>
+            <version>6.0</version>
             <scope>compile</scope>
             <optional>true</optional>
             <exclusions>
--- a/rules-java-archives/addon/src/main/java/org/jboss/windup/rules/apps/java/archives/identify/LuceneArchiveIdentificationService.java
+++ b/rules-java-archives/addon/src/main/java/org/jboss/windup/rules/apps/java/archives/identify/LuceneArchiveIdentificationService.java
@@ -1,19 +1,13 @@
 package org.jboss.windup.rules.apps.java.archives.identify;
 
 import java.io.File;
-import java.io.IOException;
 import java.util.logging.Logger;
 
 import org.apache.lucene.document.Document;
-import org.apache.lucene.index.Term;
-import org.apache.lucene.search.Query;
-import org.apache.lucene.search.ScoreDoc;
-import org.apache.lucene.search.TermQuery;
-import org.apache.lucene.search.TopDocs;
 import org.jboss.forge.addon.dependencies.Coordinate;
 import org.jboss.forge.addon.dependencies.builder.CoordinateBuilder;
+import org.jboss.windup.maven.nexusindexer.client.DocTo;
 import org.jboss.windup.util.Logging;
-import org.jboss.windup.util.exception.WindupException;
 
 /**
  * Identifies archives by their hash, using pre-created Lucene index. See the nexus-repository-indexer project.
@@ -23,7 +17,7 @@ import org.jboss.windup.util.exception.WindupException;
  *
  * TODO: This should be in Nexus Indexer - Data (client for Nexus Indexer - Core).
  */
-public class LuceneArchiveIdentificationService extends LuceneIndexServiceBase implements ArchiveIdentificationService
+public class LuceneArchiveIdentificationService extends org.jboss.windup.maven.nexusindexer.client.LuceneIndexServiceBase implements ArchiveIdentificationService
 {
     private static final Logger LOG = Logging.get(LuceneArchiveIdentificationService.class);
 
@@ -42,31 +36,19 @@ public class LuceneArchiveIdentificationService extends LuceneIndexServiceBase i
 
 
     @Override
-    public Coordinate getCoordinate(String checksum)
+    public Coordinate getCoordinate(String sha1)
     {
-        Query query = new TermQuery(new Term(SHA1, checksum));
-        try
+        return this.findSingle(DocTo.Fields.SHA1, sha1, new DocTo<Coordinate>()
         {
-            TopDocs results = searcher.search(query, 100);
-            for (ScoreDoc scoreDoc : results.scoreDocs)
+            public Coordinate convert(Document doc)
             {
-                Document doc = searcher.doc(scoreDoc.doc);
-                String groupId = doc.get(GROUP_ID);
-                String artifactId = doc.get(ARTIFACT_ID);
-                String version = doc.get(VERSION);
-                String classifier = doc.get(CLASSIFIER);
-                String packaging = doc.get(PACKAGING);
-
-                Coordinate coordinate = CoordinateBuilder.create()
-                    .setGroupId(groupId).setArtifactId(artifactId).setVersion(version)
-                    .setClassifier(classifier).setPackaging(packaging);
-                return coordinate;
+                return CoordinateBuilder.create()
+                    .setGroupId(doc.get(GROUP_ID))
+                    .setArtifactId(doc.get(ARTIFACT_ID))
+                    .setVersion(doc.get(VERSION))
+                    .setClassifier(doc.get(CLASSIFIER))
+                    .setPackaging(doc.get(PACKAGING));
             }
-            return null;
-        }
-        catch (IOException e)
-        {
-            throw new WindupException("Failed to find Maven coords for SHA1: " + checksum + " due to: " + e.getMessage(), e);
-        }
+        });
     }
 }
--- a/rules-java-archives/addon/src/main/java/org/jboss/windup/rules/apps/java/archives/identify/LuceneArchiveIdentificationService.java
+++ b/rules-java-archives/addon/src/main/java/org/jboss/windup/rules/apps/java/archives/identify/LuceneArchiveIdentificationService.java
@@ -1,24 +1,17 @@
 package org.jboss.windup.rules.apps.java.archives.identify;
 
-import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.util.logging.Logger;
 
 import org.apache.lucene.document.Document;
-import org.apache.lucene.index.DirectoryReader;
-import org.apache.lucene.index.IndexReader;
 import org.apache.lucene.index.Term;
-import org.apache.lucene.search.IndexSearcher;
 import org.apache.lucene.search.Query;
 import org.apache.lucene.search.ScoreDoc;
 import org.apache.lucene.search.TermQuery;
 import org.apache.lucene.search.TopDocs;
-import org.apache.lucene.store.Directory;
-import org.apache.lucene.store.SimpleFSDirectory;
 import org.jboss.forge.addon.dependencies.Coordinate;
 import org.jboss.forge.addon.dependencies.builder.CoordinateBuilder;
-import org.jboss.forge.furnace.util.Assert;
 import org.jboss.windup.util.Logging;
 import org.jboss.windup.util.exception.WindupException;
 
@@ -26,8 +19,11 @@ import org.jboss.windup.util.exception.WindupException;
  * Identifies archives by their hash, using pre-created Lucene index. See the nexus-repository-indexer project.
  *
  * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
+ * @author <a href="http://ondra.zizka.cz/">Ondrej Zizka, zizka@seznam.cz</a>
+ *
+ * TODO: This should be in Nexus Indexer - Data (client for Nexus Indexer - Core).
  */
-public class LuceneArchiveIdentificationService implements ArchiveIdentificationService, Closeable
+public class LuceneArchiveIdentificationService extends LuceneIndexServiceBase implements ArchiveIdentificationService
 {
     private static final Logger LOG = Logging.get(LuceneArchiveIdentificationService.class);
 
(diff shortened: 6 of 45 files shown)


jsight:
It looks like this has a lot of unrelated commits and things.

OndraZizka:
Right. I found the API we use to do Java-based rules quite inconvenient and put few things to make it easier. Should I separate that?

jsight:
No, you previously had things from another PR in here as well. I guess that was cleaned up after the rebase. :)

OndraZizka:
I've moved some of the changes to individual PR's.
#900 #899 #898 #897 + #896 #894

OndraZizka:
This is something between "ready to review" and "work in progress". Not squashed, tested, but I've added some untested code, will test in the evening. It creates the 1st stage pom.xml structure with relatively usable output, although there are many things to improve. These improvements need small refactoring of the Maven and Archive related models. Also, the package index is not used yet.

OndraZizka:
Squashed

OndraZizka:
Squashed even more.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
```diff
@@ -0,0 +1,187 @@
...
+        NOTE: Configuring the Maven repository in the pom.xml file is not a recommended procedure
+        and is only done here to make it easier.
+        See the section entitled 'Use the Maven Repository' in the Development Guide for Red Hat JBoss EAP:
+        https://access.redhat.com/documentation/en/jboss-enterprise-application-platform/
+    -->
+    <repositories>
+        <repository>
+            <id>jboss-developer-staging-repository</id>
+            <url>http://jboss-developer.github.io/temp-maven-repo/</url>
+        </repository>
```
mareknovotny: This repository should not be provided as this is old and now not used.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
```diff
@@ -0,0 +1,187 @@
...
+            <id>jboss-enterprise-maven-repository</id>
+            <url>https://maven.repository.redhat.com/ga/</url>
+        </repository>
+    </repositories>
+
+    <pluginRepositories>
+        <pluginRepository>
+            <id>jboss-developer-staging-repository</id>
+            <url>http://jboss-developer.github.io/temp-maven-repo/</url>
+        </pluginRepository>
```
mareknovotny: This repository should not be provided as this is old and now not used

rules-java/tests/pom.xml
```diff
@@ -107,6 +110,48 @@
...
+            <type>zip</type>
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.shrinkwrap.resolver</groupId>
+            <artifactId>shrinkwrap-resolver-depchain</artifactId>
+            <version>2.2.2</version>
+            <scope>test</scope>
+            <type>pom</type>
+        </dependency>
```
mareknovotny: duplicated dependency see line # 126-132

rules-java/tests/pom.xml
```diff
@@ -107,6 +110,48 @@
...
+            <version>3.3.3</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.shrinkwrap.resolver</groupId>
+            <artifactId>shrinkwrap-resolver-depchain</artifactId>
+            <version>2.2.2</version>
+            <scope>test</scope>
+            <type>pom</type>
+        </dependency>
```
mareknovotny: duplicated dependency see my comment https://github.com/windup/windup/pull/878/files#r62461207

rules-java/tests/pom.xml
```diff
@@ -12,6 +12,10 @@
 
     <name>Windup Rules - Java - Basic Tests</name>
 
+    <properties>
+        <version.nexus.index>5</version.nexus.index>
+    </properties>
+
```
mareknovotny: this is already defined in windup-bom, why should be again here?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/bom.xml.ftl
```diff
@@ -0,0 +1,61 @@
...
+
+            <!-- JBoss distributes a complete set of Java EE APIs including a Bill
+                of Materials (BOM). A BOM specifies the versions of a "stack" (or a collection)
+                of artifacts. We use this here so that we always get the correct versions
+                of artifacts. Here we use the jboss-eap-javaee7 stack (you can
+                read this as the JBoss stack of the Java EE APIs and related components.  -->
+            <dependency>
+                <groupId>org.jboss.bom</groupId>
+                <artifactId>jboss-eap-javaee7</artifactId>
+                <version>7.0.0-build-12</version>
```
mareknovotny: we should show something different than this BOM.
```
           <dependency>
                <groupId>org.jboss.spec</groupId>
                <artifactId>jboss-javaee-6.0</artifactId>
                <version>${version.org.jboss.spec.jboss.javaee.6.0}</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
```
where `version.org.jboss.spec.jboss.javaee.6.0` is something like *3.0.3.Final*

If we want to use extended BOM with JBoss GAVs not only Java EE specific, we should use:
```
           <dependency>
                <groupId>org.jboss.spec</groupId>
                <artifactId>jboss-javaee-6.0-with-all</artifactId>
                <version>1.0.7.Final</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
```
You can use instead of jboss-javaee-6.0-with-all some specific (http://search.maven.org/#search|ga|1|g%3A%22org.jboss.bom%22%20AND%20v%3A%221.0.7.Final%22), but I would not recommend it for our purpose

 At least I can see here a mismatch between EAR maven plugin definition as Java EE 6 and now we provide BOM with Java EE 7. There are similar BOMs for Java EE 7 look at http://search.maven.org/#search|gav|1|g%3A%22org.wildfly.bom%22%20AND%20a%3A%22jboss-javaee-7.0-wildfly%22

rules-java-archives/addon/pom.xml
```diff
@@ -15,12 +15,31 @@
 
         <!-- Local Dependencies -->
         <dependency>
+            <groupId>org.jboss.windup.maven</groupId>
+            <artifactId>nexus-indexer-data</artifactId>
+            <version>6-SNAPSHOT</version>
```
mareknovotny: we probably need to update the version to 6, right?

rules-java/api/pom.xml
```diff
@@ -35,6 +35,23 @@
             <version>5.0.4</version>
         </dependency>
 
+        <!-- Local Dependencies -->
+        <dependency>
+            <groupId>org.jboss.windup.maven</groupId>
+            <artifactId>nexus-indexer-data</artifactId>
+            <version>6-SNAPSHOT</version>
```
mareknovotny: whyt not dependency version from windup-bom?

rules-java/api/pom.xml
```diff
@@ -35,6 +35,23 @@
...
+            <scope>compile</scope>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-core</artifactId>
+            <version>4.8.1</version>
```
mareknovotny: whyt not dependency defined in windup-bom?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenCoord.java
```diff
@@ -0,0 +1,237 @@
...
+    public MavenCoord setClassifier(String classifier)
+    {
+        this.classifier = classifier;
+        return this;
+    }
+
+
+    public String getPackaging()
+    {
+        return packaging; // == null ? "pom" : packaging;
```
mareknovotny: leftover after code change?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizationService.java
```diff
@@ -0,0 +1,384 @@
...
+        public GraphContext getGraphContext() {
+            return graphContext;
+        }
+
+    }
+
+    /**
+     * This will be used to determine which BOM to take.
+     * Currently we only have these tags: eap, eap7.
+     */
```
mareknovotny: So this code already takes in mind the different java ee version. So i think we need to parametrize the BOMs specifically to this (javaee-6 ->BOM and javaee-7 -> BOM as I commented in BOM template)

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizeRuleProvider.java
```diff
@@ -0,0 +1,153 @@
...
+            .when(Query.fromType(IdentifiedArchiveModel.class))
+            .perform(new MavenizePutNewerVersionToGlobalBomOperation())
+            .withId("Mavenize-BOM-file-creation")
+
+            // For each application given to Windup as input, mavenize it.
+            .addRule()
+            .when(applicationProjectModels, SourceMode.isDisabled())
+            .perform(new MavenizeApplicationOperation())
+            .withId("Mavenize-projects-mavenization")
+            /**/;
```
mareknovotny: leftover of something?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/ModuleAnalysisHelper.java
```diff
@@ -0,0 +1,149 @@
...
+    String deriveGroupIdFromPackages(ProjectModel projectModel)
+    {
+        Map<String,Integer> pkgsMap = new HashMap<>();
+        Set<String> pkgs = new HashSet<>(1000);
+        GremlinPipeline<Vertex, Vertex> pipeline = new GremlinPipeline<>(projectModel);
+
+        PipeFunction<String, String> upToThirdDot =
+                new PipeFunction<String, String>(){
+                    @Override public String compute(String pkgName) {
+                        int upToThirdDot = StringUtils.ordinalIndexOf(pkgName, ".", 3);
```
mareknovotny: Is this safe to assume groupId is always 3-dots from left?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
```diff
@@ -0,0 +1,187 @@
...
+    <#if pom.coords.classifier??>${i1}<classifier>${pom.coords.classifier}</classifier>${"\n"}</#if><#t>
+
+    <#if pom.name??>${i1}<name>${pom.name}</name>${"\n"}</#if><#t>
+    <#if pom.description??>${i1}<description>${pom.description}</description>${"\n"}</#if><#t>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <!-- maven-compiler-plugin -->
+        <maven.compiler.target>1.8</maven.compiler.target>
+        <maven.compiler.source>1.8</maven.compiler.source>
```
mareknovotny: Do we assume only java 8 is the target here? 

Do not raise a problem, but we perhaps could guess the version from a runtime java?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
```diff
@@ -0,0 +1,187 @@
...
+        </pluginRepository>
+    </pluginRepositories>
+
+    <build>
+        <plugins>
+            <!-- The WildFly plug-in deploys your EAR to a local JBoss EAP container. -->
+            <plugin>
+                <groupId>org.wildfly.plugins</groupId>
+                <artifactId>wildfly-maven-plugin</artifactId>
+                <version>1.0.2.Final</version>
```
mareknovotny: shouldn't we use a property for this to get easier to change it on top of the pom.xml?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
```diff
@@ -0,0 +1,187 @@
...
+    </#if>
+    <#if pom.coords.packaging = "ear">
+    <build>
+        <finalName>${"$"}{project.parent.artifactId}</finalName>
+        <plugins>
+            <!-- EAR plug-in -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-ear-plugin</artifactId>
+                <version>2.10.1</version>
```
mareknovotny: shouldn't we use a property for this to get easier to change it on top of the pom.xml?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
OndraZizka: I took this from EAP quickstarts...

rules-java/api/pom.xml
jsight: We don't put external dependencies like that into the bom. Although it does seem like this could be pulled in from graph-api which already exposes lucene capabilities to the rest of Windup.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
OndraZizka: Fixed

rules-java/tests/pom.xml
OndraZizka: Result of rebasing... thanks, fixed

rules-java/tests/pom.xml
OndraZizka: Fixed

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/ModuleAnalysisHelper.java
jsight: Honestly, I think I would prefer imperative code over using a GremlinPipe here. The pipe is really hard to read here to me and I'm not seeing what it adds in this scenario. That's not a show-stopper, though, IMO.

What happens if there are fewer than 3 dots? It seems to me that a more reasonable default could be derived from the archive name.

rules-java/tests/pom.xml
OndraZizka: Not sure, removed

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/bom.xml.ftl
OndraZizka: Again, this is taken from EAP quickstarts. Maybe we should fix those first and I'll follow the changes here?

OndraZizka: As we looked up, the current GAV for EAP 7 is  org.jboss.bom:wildfly-javaee7
See https://github.com/wildfly/boms/commit/315698ed776fd32230ade0d537c23f6344c43e8b
And http://search.maven.org/#search|ga|1|g%3A%22org.wildfly.bom%22

rules-java-archives/addon/pom.xml
OndraZizka: Because we have a circular dependency here.

rules-java/api/pom.xml
OndraZizka: Set to 6.

rules-java/api/pom.xml
OndraZizka: Maybe we could create some kind of windup-deps-bom. IDK how much sense it makes for modules, though.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenCoord.java
OndraZizka: Removed

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizationService.java
OndraZizka: Right... it's not that hard actually, I'll implement it.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizeRuleProvider.java
OndraZizka: That's how I disable parts of code when debugging :) Removed

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/ModuleAnalysisHelper.java
OndraZizka: This is just a helper, it isn't something really hard-setting the package. It's something to use if the user didn't specify using --mavenize.groupId, and the archive or project name is something insane, like few sencences paragraph (a description) or a number or such.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
OndraZizka: Perhaps? But IDK if we can assume that what they asses the old app with will be the runtime env of the migrated application :) We can parametrize. Let's change this if anyone asks for that, but MarcZ specifically said that basically he only needs correct dependencies, and the rest are 2nd stage things.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
OndraZizka: This is only set at one place in the whole project. So I'd let it be this way... unless users demand other way. We can ask.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
OndraZizka: Here it would make more sense, as it appears for each ear... IDK. But we can keep it up-to-date in the template. Whether users will change this often, I'm not sure.

OndraZizka:
I've fixed some of the things from PR comments. Not all though - I want to keep this limited to what is the initial feature request. When it's merged, I'm going to polish it (move version to properties, etc.).

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
mareknovotny: it could be done in the root of the project in `pluginManagement` section and then all submodules will inherit versions and config. But no big deal if that isn't fixed t right now ;)

rules-java/api/pom.xml
mareknovotny: we don't but we definitely could/should it is better to have it defined there and just align on one version if there is not the issues with major versions of such deps cross submodules.

mareknovotny:
it seems much better now.

rules-java-archives/addon/pom.xml
```diff
@@ -15,12 +15,31 @@
 
         <!-- Local Dependencies -->
         <dependency>
+            <groupId>org.jboss.windup.maven</groupId>
+            <artifactId>nexus-indexer-data</artifactId>
+            <version>6.0</version>
+            <scope>compile</scope>
+            <optional>true</optional>
```
mareknovotny: so this still stays without definition in windup-bom, this is not external dependency in the reality, right ?

rules-java/api/pom.xml
```diff
@@ -35,6 +35,23 @@
...
         </dependency>
 
+        <!-- Local Dependencies -->
+        <dependency>
+            <groupId>org.jboss.windup.maven</groupId>
+            <artifactId>nexus-indexer-data</artifactId>
+            <version>6.0</version>
+            <scope>compile</scope>
+            <optional>true</optional>
+        </dependency>
```
mareknovotny: so this still stays without definition in windup-bom, this is not external dependency in the reality, right ?

rules-java-archives/addon/pom.xml
jsight: Why is there a circular dependency?

rules-java/api/pom.xml
jsight: A dependency only add-on is better than using a bom for something like this. That way we can deploy the dependency to a single place.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
jsight: I'd suggest just defaulting to 1.7 for now. But I agree that this isn't particularly important for the typical use case of these POMs. Some hand tweaking is expected.

rules-java-archives/addon/pom.xml
OndraZizka: Not sure what you mean by external. It's something this needs to bring along, but not the transitive dependencies, hence <optional>, I think.

rules-java/api/pom.xml
OndraZizka: Not sure what you mean by external. It's something this needs to bring along, but not the transitive dependencies, hence , I think.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/pom.xml.ftl
OndraZizka: I'll do it right at next iteration.

OndraZizka:
I will add explanations to TODOs. Typically I write TODO: when totally tired and have no energy to actually implement it :)

OndraZizka:
TODO's sanitized.

jsight:
retest

config/api/src/main/java/org/jboss/windup/config/query/QueryTypeCriterion.java
```diff
@@ -19,7 +19,7 @@
     public QueryTypeCriterion(Class<? extends WindupVertexFrame> clazz)
     {
         this.searchedClass = clazz;
-        this.typeValue = getTypeValue(clazz);
+        this.typeValue = TypeAwareFramedGraphQuery.getTypeValue(clazz);
```
jsight: Good :)

config/api/src/main/java/org/jboss/windup/config/query/QueryTypeCriterion.java
```diff
@@ -28,26 +28,14 @@ public void query(FramedGraphQuery q)
...
-    }
 
     /**
      * Adds a criterion to given pipeline which filters out vertices representing given WindupVertexFrame.
      */
     public static GremlinPipeline<Vertex, Vertex> addPipeFor(GremlinPipeline<Vertex, Vertex> pipeline,
                 Class<? extends WindupVertexFrame> clazz)
     {
-        pipeline.has(WindupVertexFrame.TYPE_PROP, getTypeValue(clazz));
+        pipeline.has(WindupVertexFrame.TYPE_PROP, TypeAwareFramedGraphQuery.getTypeValue(clazz));
```
jsight: Good

decompiler/impl-fernflower/src/main/java/org/jboss/windup/decompiler/fernflower/FernflowerDecompiler.java
```diff
@@ -165,7 +158,8 @@ public void decompilationProcessComplete()
 
 
 
-    private List<String> pathsFromDecompilationRequests(List<ClassDecompileRequest> requests) {
+    private List<String> pathsFromDecompilationRequests(List<ClassDecompileRequest> requests)
```
jsight: Good :)

graph/api/src/main/java/org/jboss/windup/graph/GraphContext.java
```diff
@@ -68,7 +68,12 @@
     void setOptions(Map<String, Object> options);
 
     /**
-     * Returns the globally configured options as an immutable {@link Map}
+     * Returns the globally configured options as an immutable {@link Map}.
+     *
+     * Example usage:
```
jsight: Nice!

reporting/api/src/main/java/org/jboss/windup/reporting/freemarker/FurnaceFreeMarkerTemplateLoader.java
```diff
@@ -11,7 +11,7 @@
 
 import freemarker.cache.URLTemplateLoader;
 
-class FurnaceFreeMarkerTemplateLoader extends URLTemplateLoader
+public class FurnaceFreeMarkerTemplateLoader extends URLTemplateLoader
```
jsight: Good idea to reuse this.

rules-java-archives/addon/src/main/java/org/jboss/windup/rules/apps/java/archives/identify/LuceneArchiveIdentificationService.java
```diff
@@ -38,68 +28,27 @@
...
+    public Coordinate getCoordinate(String sha1)
     {
-
-        Query query = new TermQuery(new Term("sha1", checksum));
-        try
+        return this.findSingle(DocTo.Fields.SHA1, sha1, new DocTo<Coordinate>()
         {
-            TopDocs results = searcher.search(query, 100);
-            for (ScoreDoc scoreDoc : results.scoreDocs)
+            public Coordinate convert(Document doc)
```
jsight: THis seems mushc simpler. Good.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/ApiDependenciesData.java
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * This class was originally intended as a static store of various BOM's dependencies.
+ * But it turned out there are so many that it's better to extract them right from the BOM.
+ * That is implemented in nexus-repo-indexer 6.
+ *
+ * This might be deleted in the final version.
+ *
+ * @author <a href="http://ondra.zizka.cz/">Ondrej Zizka, zizka@seznam.cz</a>
+ */
+interface ApiDependenciesData
```
jsight: I look forward to this disappearing. :)

OndraZizka: Yeah... I thought it would ony be like 10 things but then I saw that each technology has around 5 artefacts... and there are dozens of techs.

decompiler/impl-fernflower/src/main/java/org/jboss/windup/decompiler/fernflower/FernflowerDecompiler.java
OndraZizka: Formatting master :)

rules-java/api/pom.xml
OndraZizka: I agree that we should better keep things isolated to addons.
If it happens that same dependency appears in more than, say, 5 addons, we could create another addon just for that dependency.
BOM may define those which go throughout the project, like commons-lang.
See also WINDUP-1007.

jsight:
retest

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizationService.java
```diff
@@ -0,0 +1,422 @@
...
+                    LOG.info("Library from 3rd party vendor ("+org.getName()+"), skipping recursive mavenization: " + arch.getFilePath());
+                    return null;
+                }
+            }
+        }
+
+        // A MavenProject.
+        // TODO: This covers both app modules and libraries with pom.xml inside. Need to diferentiate - skip only libraries.
+        // TODO: This should disappear after WINDUP-981
+        if (false && projectModel instanceof MavenProjectModel)
```
mareknovotny: Eclipse sees this as dead code. I read TODO, but don't know why it still here ;)

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizationService.java
```diff
@@ -0,0 +1,422 @@
...
+    {
+        Set<MavenCoord> dependenciesMet = new HashSet();
+        dependenciesMet.add(pom.coord);
+
+        SortedSet<MavenCoord> dependenciesSatisfied = new TreeSet<>();
+
+        for (Dependency dep : pom.dependencies)
+        {
+            // Traverse the tree, depth-first, take items at node exit.
+        }
```
mareknovotny: Shouldn't this for cycle be commented until an operation is created?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizationService.java
```diff
@@ -0,0 +1,422 @@
...
+    /**
+     * Sorts the submodules of given Pom so that their cross-dependencies are satisfied if built in that order.
+     * TODO - MIGR-236.
+     */
+    private OrderedMap<String, Pom> sortSubmodulesToReflectDependencies(Pom pom)
+    {
+        Set<MavenCoord> dependenciesMet = new HashSet();
+        dependenciesMet.add(pom.coord);
+
+        SortedSet<MavenCoord> dependenciesSatisfied = new TreeSet<>();
```
mareknovotny: variable not used anywhere

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/MavenizationService.java
```diff
@@ -0,0 +1,422 @@
...
+            modulePom.dependencies.add(subModulePom);
+        }
+
+        // Nested module candidates.
+        for (ArchiveModel nestedModule : nestedModules)
+        {
+            // TODO: Is it a submodule or a library? Does it appear in multiple applications?
+            //Pom subModulePom = mavenizeModule(mavCtx, nestedModule, containingModule);
+            //modulePom.dependencies.add(subModulePom.identification);
+        }
```
mareknovotny: Shouldn't this for cycle be commented until an operation is created?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/PackagesToContainingMavenArtifactsIndex.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * After the packages are registered and Java scanning done,
+     * we can link the ProjectModel and API packages together.
+     * ProjectModel --uses--> ArchiveCoordinateModel
+     */
+    public void markProjectsUsingPackagesFromAPI(MavenCoord apiCoords)
+    {
+        final GraphService<ArchiveCoordinateModel> coordsService = graphContext.service(ArchiveCoordinateModel.class);
```
mareknovotny: variable is not used anywhere.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/GlobalBomModel.java
```diff
@@ -0,0 +1,69 @@
...
+    @Adjacency(label = DEFINES, direction = Direction.OUT)
+    void addNewerDependency(ArchiveCoordinateModel dependency);
+
+    public abstract class Impl implements GlobalBomModel, JavaHandlerContext<Vertex> {
+        private static final Logger LOG = Logging.get(MavenizeRuleProvider.class);
+
+        @Override
+        public void addNewerDependency(final ArchiveCoordinateModel newCoord){
+            LOG.info("Adding: " + newCoord.toString());
+            Vertex v = this.it();
```
mareknovotny: variable is not used anywhere

rules-java/api/src/main/java/org/jboss/windup/rules/apps/mavenize/FeatureBasedApiDependenciesDeducer.java
```diff
@@ -0,0 +1,113 @@
...
+    private boolean hasJaxrs(ProjectModel projectModel)
+    {
+        return false;
+    }
+
+
+    private boolean moduleContainsClassesFromAPI(ProjectModel projectModel, MavenCoord apiCoords)
+    {
+        throw new UnsupportedOperationException("Not supported yet."); //To change body of generated methods, choose Tools | Templates.
+    }
```
mareknovotny: Is this generated method left over as it is not used anywhere and not forced to override from parent/inheritance

PR #881, opened 2016-04-20
WINDUP-881: FreeMarker template error on an ejb-jar.xml that doesn't …

jsight:
https://issues.jboss.org/browse/WINDUP-881

### DIFF
--- a/rules-java-ee/addon/src/main/resources/reports/templates/jboss/jboss-ejb3.ftl
+++ b/rules-java-ee/addon/src/main/resources/reports/templates/jboss/jboss-ejb3.ftl
@@ -81,7 +81,7 @@
         <#list reportModel.relatedResources.sessionBeans.list.iterator() as sessionBean>
         <session>
             <ejb-name>${sessionBean.beanName}</ejb-name>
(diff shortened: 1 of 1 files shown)


rules-java-ee/addon/src/main/resources/reports/templates/jboss/jboss-ejb3.ftl
```diff
@@ -81,7 +81,7 @@
         <#list reportModel.relatedResources.sessionBeans.list.iterator() as sessionBean>
         <session>
             <ejb-name>${sessionBean.beanName}</ejb-name>
-            <session-type>${sessionBean.sessionType}</session-type>
+            <#if sessionBean.sessionType??><session-type>${sessionBean.sessionType}</session-type></#if>
```
OndraZizka: FYI - https://issues.apache.org/jira/browse/FREEMARKER-21

jsight: Yes, that new syntax would be nice.

PR #882, opened 2016-04-20
WINDUP-965: Windup will suggest loading CR versions of rules even if …

jsight:
https://issues.jboss.org/browse/WINDUP-965

### DIFF
--- a/exec/api/src/main/java/org/jboss/windup/exec/updater/RulesetsUpdater.java
+++ b/exec/api/src/main/java/org/jboss/windup/exec/updater/RulesetsUpdater.java
@@ -7,6 +7,7 @@ import java.util.List;
 import java.util.logging.Logger;
 import javax.inject.Inject;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang.StringUtils;
 import org.jboss.forge.addon.dependencies.Coordinate;
 import org.jboss.forge.addon.dependencies.Dependency;
 import org.jboss.forge.addon.dependencies.DependencyException;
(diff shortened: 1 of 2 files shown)


exec/api/src/main/java/org/jboss/windup/exec/updater/RulesetsUpdater.java
```diff
@@ -123,10 +124,13 @@ public Coordinate getLatestReleaseOf(final CoordinateBuilder coord)
...
         List<Coordinate> availableVersions = depsResolver.resolveVersions(DependencyQueryBuilder.create(coord));
 
-        // Find the latest non-SNAPSHOT version.
+        // Find the latest non-SNAPSHOT and non-CR version.
         for(int i = availableVersions.size()-1; i >= 0; i--)
         {
-            if(!availableVersions.get(i).isSnapshot())
+            Coordinate availableCoord = availableVersions.get(i);
+            String versionStr = availableCoord.getVersion() == null ? "" : availableCoord.getVersion();
+
```
OndraZizka: What's the case where version would be null?

exec/api/src/main/java/org/jboss/windup/exec/updater/RulesetsUpdater.java
```diff
@@ -123,10 +124,13 @@ public Coordinate getLatestReleaseOf(final CoordinateBuilder coord)
...
+            if(!availableCoord.isSnapshot() && !versionStr.matches(".*CR[0-9]$"))
                 return availableVersions.get(i);
```
OndraZizka: If the version is null, this will return it. Is that ok?

exec/api/src/main/java/org/jboss/windup/exec/updater/RulesetsUpdater.java
jsight: I don't think it will ever be null. That code is probably overly paranoid.

jsight:
Tweaked it a bit based on your feedback. I don't think returning one with a null version would make sense (if that ever happens).

PR #887, opened 2016-04-21
WINDUP-935 Shorten the label that breaks tables and charts.

OndraZizka:


jsight:
This will need approval from Tobias.

mareknovotny:
has somebody already asked Tobias or others if they are fine with the change? 
IMHO this is good to go as it just unifies the labels in one word and description in more words as it should be.
@Maarc could you confirm this is good generally?

thartwig-rht:
Good for me!

PR #891, opened 2016-04-22
[WINDUP-992] check classname against the list of scanned packages correctly

Ladicek:


jsight:
Great catch.. thank you.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/scan/operation/packagemapping/PackageNameMapping.java
```diff
@@ -14,6 +14,7 @@
 import org.jboss.windup.config.GraphRewrite;
 import org.jboss.windup.config.GraphRule;
 import org.jboss.windup.config.PreRulesetEvaluation;
+import org.jboss.windup.rules.apps.java.model.PackageModel;
```
Ladicek: Wow, did I add an unused import during my experiments? Probably... :-(

jsight: No worries... will fix in my next PR.

PR #895, opened 2016-04-25
Used update instead of the previous single row approach

jsight:
Amends https://github.com/windup/windup/pull/889 with a better fix.


OndraZizka:
Works for me. Is this better than the previous?

jsight:
@OndraZizka - I think that it looks cleaner.

PR #898, opened 2016-04-25
Javadoc improvement

OndraZizka:
From https://github.com/windup/windup/pull/878/commits/24a5de4912dc874fb55fef16a11b98c16703e1f3


jsight:
Ok to merge, but in general it is better to integrate these types of things with larger feature PRs.

OndraZizka:
Well, maybe? But then you say "It looks like this has a lot of unrelated commits and things"... so I'm not sure which one to pick? :)

jsight:
@OndraZizka - That is what I said in the the case of PRs containing entire copies of other PRs. :)

Comment changes and general small refactorings are fine to be combined with feature work.

PR #899, opened 2016-04-25
Archive identification listener - print the SHA1.

OndraZizka:


rules-java-archives/addon/src/main/java/org/jboss/windup/rules/apps/java/archives/listener/ArchiveIdentificationGraphChangedListener.java
```diff
@@ -21,43 +21,49 @@
...
 {
-    private static final Logger log = Logger.getLogger(ArchiveIdentificationGraphChangedListener.class.getName());
+    private static final Logger LOG = Logging.get(ArchiveIdentificationGraphChangedListener.class);
 
     private final ArchiveIdentificationService identifier;
-
     private GraphContext context;
+    private ArchiveService archiveService;
+
+    public ArchiveIdentificationGraphChangedListener setGraphContext(GraphContext context)
```
jsight: Why is this public?

OndraZizka: Sometimes the might want to set the graph context? Otherwise I think no reason rather than IDE generating it that way by default.

PR #904, opened 2016-04-27
Fixed package filtering during class file prescanning

jsight:


OndraZizka:
I'm not sure if removing decomposition is a good idea, but better have it working than decomposed :)

jsight:
@OndraZizka - What do you mean regarding removing decomposition? The only class that I removed was not used before at all (it was dead code).

OndraZizka:
Right, I didn't notice...

jsight:
@OndraZizka - Understood :) I only saw it because I was searching for any similar code that may have had similarly broken assumptions.

PR #906, opened 2016-04-29
make app list title/text style consistent with other reports

d-s:
In the title and help text for other reports: all words capitalized in title, word "report" not capitalized in help text


windupgithubbot1:

PR #907, opened 2016-04-30
WINDUP-1008 Improve the layout of Application List

OndraZizka:
See jira for screenshots.


OndraZizka:
retest

OndraZizka:
This will probably need an update of tests.

mareknovotny:
I think al such layout changes requires confirmation from the our team not only committers ;)

OndraZizka:
retest

jsight:
+1 - This is actually a major change to the layout that will need more extensive review.

OndraZizka:
In my opinion, there's no doubt this is better. But who knows, maybe someone will like the table layout for non tabular data more, until someone from UX tells it's really not nice ;-)

jsight:
If nothing else, I really dislike the little x's all over the numbers.

OndraZizka:
That indicates that it's not story points. Imo just numbers is pretty confusing and we heard that many times.

d-s:
Agreed that we shouldn't just cram this PR in without review/discussion,
but I do think it looks really nice -- and more in sync with the
application details report.

(I don't love the Xs myself, but the charts in the application details
report already present incident counts the same way, and I can't really
think of a better way to communicate incident counts vs. story points...)

mareknovotny:
I think times symbol is good to imagine on first view what it is, and the columns really looks more like overview which is good. 
I haven't commented on order of columns as that is only disputable what should be in first, second etc. and if we want to provide sorting. I think this will be the major point for discussion with Robb and Tobias.

OndraZizka:
Squashed

OndraZizka:
Ok so I will update the tests and merge this, as agreed on Monday's meeting.

OndraZizka:
Also, it got broken over time, need to fix the CSS.

jsight:
retest

OndraZizka:
retest

PR #908, opened 2016-04-30
    WINDUP-987 Trim the text of titles, hints and classifications

OndraZizka:
This may sound trivial, but it will reduce the size of the reports .html and .js by few megabytes.
The values need to be trimmed right when created, not in the reports, because somewhere they are printed through things like ObjectWrapper.


jsight:
retest

mareknovotny:
I guess this is ready for a review ;)

jsight:
I doubt this will even save one megabyte, but I guess its ok. :)

OndraZizka:
@jsight It saves several megabytes for our biggest app just for the data .js file.

jsight:
@OndraZizka - Interesting and good to know. :) Thanks.

PR #909, opened 2016-04-30
WINDUP-1009 Replace commons-lang 2.x with 3.x

OndraZizka:
Also related to WINDUP-1007 Review the flat dependencies of windup-distribution

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -12,7 +12,7 @@
     <parent>
         <groupId>org.jboss</groupId>
         <artifactId>jboss-parent</artifactId>
-        <version>19</version>
+        <version>20</version>
         <relativePath />
     </parent>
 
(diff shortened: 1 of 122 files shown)


jsight:
@windupgithubbot1 retest

OndraZizka:
retest

mareknovotny:
@windupgithubbot1 retest

mareknovotny:
It seems that jenkins is stalled or so, I ran tests locally and all passed.

jsight:
After this change, commons-logging 3 is now deployed in three new places within windup:
addons/org-jboss-windup-rules-apps-windup-rules-java-ee-2-5-0-SNAPSHOT/commons-lang3-3.4.jar
addons/org-jboss-windup-graph-windup-graph-2-5-0-SNAPSHOT/commons-lang3-3.4.jar
addons/org-jboss-windup-rules-apps-windup-rules-java-2-5-0-SNAPSHOT/commons-lang3-3.4.jar

That seems strange to me and makes think that the dependencies aren't fully thought out here.

OndraZizka:
retest

OndraZizka:
@jsight   That's what I was talkinb about at the meeting - the dependencies are wrong in master, not in this PR. The commons-lang3 is used at those places because some other dependency has it. That's wrong, IMO. And we do that a lot BTW.

jsight:
Including it in 3 extra addons is actually worse, though.

OndraZizka:
How is it worse? If we removed the deps from the addons, it would fail to compile.
Should I move the commons-lang to windup-utils and depend on that then?

jsight:
It is worse as you are now classloading commons-utils 3 more times than before.

OndraZizka:
That's a fair point, but I thought that's a sacrifice you give when using modules :) So what do you suggest?

jsight:
Previously, I think that graph was exposing commons-lang. Now perhaps windup-utils could do it instead (as you were suggesting)?

OndraZizka:
okok

OndraZizka:
Changed as per comments above, and rebased. -> tests

jsight:
Ok, I will look again soon.

jsight:
retest

jsight:
retest

jsight:
retest

PR #913, opened 2016-05-03
WINDUP-1006 adding images and correct jquery-ui.min.css

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1006


OndraZizka:
I wonder what the ui-bg_flat_* images are good for rather than having a background... But I rely on jquery community to know that :) Good to go IMO

PR #921, opened 2016-05-06
Extract generic Lucene querying code into LuceneIndexServiceBase.

OndraZizka:
The plan is to move that to nexus-indexer-data, see https://github.com/windup/nexus-repository-indexer/pull/5


OndraZizka:
Part of #878

PR #927, opened 2016-05-16
WINDUP-1002 new Type location Enum

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1002

I added ENUM_CONSTANT type. It resolves/creates a ClassReference  with qualifiedName as Enum.CONSTANT


mareknovotny:
What is required to have Enum.CONSTANT searchable? I am a little lost where to add this in a RuleProvider.

java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
```diff
@@ -193,6 +195,68 @@ private String extractDefinitionLine(String typeDeclaration)
...
+     * @param clazzName - class for an Enum
+     * @param line - line with referenced Enum constant
+     * @return Fully referenced Enum constant
+     */
+    private static String extractEnumConstantFromString(String clazzName, String line)
+    {
+        Class clazz = null;
+        try
+        {
+            clazz = Class.forName(clazzName);
```
jsight: I would expect this to fail most of the time. Most of the time the code from the client's application will not be on the classpath.

jsight:
It would just need to be added to the schema, I think. org.jboss.windup.rules.apps.java.xml.TypeReferenceLocationHandler should be able to load it based upon the enum.

java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
mareknovotny: yep, that's true unfortunately. The other way of guessing constant would be probably more hard or not effective.

jsight: In the case of assignment, it is possible to get it from JDT (expression.getName(), I think). I'm not sure about the other cases.

mareknovotny:
I think that is handled already in `return Enum.valueOf(TypeReferenceLocation.class, location.trim());` at https://github.com/windup/windup/blob/master/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/xml/TypeReferenceLocationHandler.java#L41

java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
jsight: It looks like MethodInvocation has a List accessible via "arguments()" with this information as well.

mareknovotny:
The Enum const location type reference is added and both cases when Enum class is on classpath or is not it is recognized by Windup AST. The cases are method parameters, variable initialization, field declaration and constructor call.

Adding more tests to test the whole thing

forks/jdt/pom.xml
```diff
@@ -19,7 +19,7 @@
             <groupId>org.eclipse.tycho</groupId>
             <artifactId>org.eclipse.jdt.core</artifactId>
             <version>3.10.0.v20140604-1726</version>
-            <scope>provided</scope>
+<!--             <scope>provided</scope> -->
```
jsight: This will have the side effect of pulling this into the windup distribution again, creating a conflict between our modified version and the upstream version.

I don't think we can leave it this way for the merge, although maybe it is your intention to remove it before merging?

utils/src/main/java/org/jboss/windup/util/ClassNameUtil.java
```diff
@@ -12,4 +13,24 @@ public static final String getPackageName(String qualifiedName)
             packageName = qualifiedName.substring(0, qualifiedName.lastIndexOf("."));
         return packageName;
     }
+    
+    /*
+     * This method assumes constants are upper cased
+     */
+    public static boolean isConstant(String fqcn)
+    {
+        return isUpperCase(fqcn);
```
jsight: Minor, but could you use org.apache.commons.lang3.StringUtils#isAllUpperCase instead?

java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
```diff
@@ -455,48 +594,102 @@ public boolean visit(FieldDeclaration node)
...
-
                     PackageAndClassName packageAndClassName = PackageAndClassName.parseFromQualifiedName(result.result);
 
+                    if (packageAndClassName.constantName != null && packageAndClassName.packageName == null)
+                    {
+                        // the Enum Class and Constant doesn't have to be FQCN as package could miss
+                        packageAndClassName = PackageAndClassName.parseFromQualifiedName(state.getClassNameToFQCN().get(packageAndClassName.className) + "."+packageAndClassName.constantName);
+
+                        //additionally add Enum Constant type reference if found in String
+                        ClassReference reference = new ClassReference(packageAndClassName.toString(),
```
jsight: How do you know this is an ENUM and not some other kind of constant?

jsight:
I've sent a PR to this branch: https://github.com/mareknovotny/windup/pull/1

forks/jdt/pom.xml
mareknovotny: +1, I had hard time to have it in provided scope for Eclipse development and Junit test running. So I have commented that line temporary.

utils/src/main/java/org/jboss/windup/util/ClassNameUtil.java
mareknovotny: no, as Apache StringUtils does not evaluate properly non letters like undescore or dashes, so it would not help for MAX_CONSTANT only for MAXCONSTANT.

java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
mareknovotny: I originally was thinking about having a type.isEnum() test, but I reworked the impl and forgot to test it for that? Good catch!

mareknovotny:
ok, i will look what needs to be changed, thanks @jsight !

mareknovotny:
it seems that for some tests it can't parse the java from decompiled bytecode. I will look into root cause failures.

java-ast/addon/src/main/java/org/jboss/windup/ast/java/ReferenceResolvingVisitor.java
```diff
@@ -1361,7 +1463,20 @@ public PackageAndClassName(String packageName, String className)
...
+            }
+            if (this.className != null)
+                sb.append(this.className);
 
+            return sb.toString();
+        }
+        
         public static PackageAndClassName parseFromQualifiedName(String qualifiedName)
         {
             final String packageName;
```
OndraZizka: (Not related to the PR) This class seriously needs to be refactored to smaller parts.

jsight: Yeah, probably. More importantly, I would like for it output a graph of type reference information rather than the current flat list.

java-ast/tests/src/test/java/org/jboss/windup/ast/java/test/JavaEnumerationScanningTest.java
```diff
@@ -0,0 +1,95 @@
...
+                ResolutionStatus.RESOLVED, TypeReferenceLocation.ENUM_CONSTANT, 31, 41, 21, "new EnumerationClassUsage(AccessMode.READ)");
+        Assert.assertTrue(references.contains(referenceTest));
+    }
+    
+    @Test
+    public void testEnumConstWithoutClassOnClasspath()
+    {
+        Set<String> libraryPaths = new HashSet<>();
+        libraryPaths.add("src/test/resources/testclasses/enumeration/dependency/hibernate-search-engine-5.5.3.Final.jar");
+
```
OndraZizka: Could we perhaps reduce the jar so we don't inflate the repo size too much?

jsight: Yeah, if we ever get around to writing our jar reducer utility, we should use it on this. In the mean time, I'm not worried about the 1 MB.

OndraZizka: 1 MB here, 1 MB there... :)

OndraZizka: Also it would be nice to have a XML test.

jsight: The java-ast module doesn't need an "XML test". :)

PR #928, opened 2016-05-17
WINDUPRULE-166 replacing Static IP with Hard-coded IP

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-166

the question is if the affected "static IP" classes should not be renamed too, but that is just about clean code  thing


jsight:
I think renaming the classes would be a good idea as well.

PR #931, opened 2016-05-17
WINDUP-1015: <description> value taken instead of <name> to create Pr…

jsight:
https://issues.jboss.org/browse/WINDUP-1015


OndraZizka:
Great! :)

OndraZizka:
Where was the value being taken from? MANIFEST? In that case, shouldn't we still have a look what's taken from MANIFEST? If it was taking Specification-Title: and that's often a long description, maybe we should use some other value.

jsight:
I don't think that it is obvious that Specification-TItle is any worse or better than Implementation-Title. It seems like Maven builds stuff the description there, but with Maven builds we have the POM to pull from.

With manually put together manifests (or maybe other build systems?), often the implementation-title will be garbage (just a package name or something) when the specification-title is good. Its really just a guess at that point.

OndraZizka:
MANIFEST "spec": https://docs.oracle.com/javase/tutorial/deployment/jar/packageman.html

OndraZizka:
For mavenization, Impl-Title might be better

PR #935, opened 2016-05-18
WINDUP-997: spring-mvc.xml as invalid by 'XML File is not valid' rule

jsight:
https://issues.jboss.org/browse/WINDUP-997

### DIFF
--- a/rules-xml/impl/src/main/java/org/jboss/windup/rules/apps/xml/xml/ValidateXmlFilesRuleProvider.java
+++ b/rules-xml/impl/src/main/java/org/jboss/windup/rules/apps/xml/xml/ValidateXmlFilesRuleProvider.java
@@ -1,16 +1,33 @@
 package org.jboss.windup.rules.apps.xml.xml;
 
+import java.io.FileNotFoundException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.UnknownHostException;
+import java.util.Collections;
+
+import javax.xml.XMLConstants;
+import javax.xml.transform.Source;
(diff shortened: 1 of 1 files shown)


rules-xml/impl/src/main/java/org/jboss/windup/rules/apps/xml/xml/ValidateXmlFilesRuleProvider.java
```diff
@@ -141,35 +152,72 @@ private FileLocationModel createLocationModelFromNodeElement(Node node, GraphCon
...
-                return false;
+                parseFailed = true;
+            }
+
+            if (parseFailed)
+            {
+                sourceFile.setGenerateSourceReport(true);
+
+                ClassificationService classificationService = new ClassificationService(event.getGraphContext());
+                ClassificationModel model =classificationService.attachClassification(context, sourceFile, XmlFileModel.NOT_VALID_XML,
```
OndraZizka: Missing space :)

OndraZizka: Missing space :)

OndraZizka:
We have the Unparsable files report, does this count with it? I know that if it doesn't, the original did not either, but once on it, we should add it to the unparsables.
The way to mark it so is to add set the `parseError` property.

jsight:
I don't think we have generally treated a xsd validation error as "unparsable". In many cases these files are actually valid and working within their context, which isn't really the same as unparsable.

PR #946, opened 2016-05-31
WINDUP-1035 Exclude viz.js from resources, save 1.3 MB 

OndraZizka:
I was curious how the graphs look like. It works with this update.


reporting/api/src/main/java/org/jboss/windup/reporting/service/ReportService.java
```diff
@@ -18,11 +18,10 @@
 
 /**
  * Convenient search and creation methods for {@link ReportModel}.
- *
  */
 public class ReportService extends GraphService<ReportModel>
 {
-    private static final String REPORTS_DIR = "reports";
+    public static final String REPORTS_DIR = "reports";
```
jsight: Use org.jboss.windup.reporting.service.ReportService#getReportDirectory for this instead of making this public.

reporting/impl/pom.xml
```diff
@@ -11,6 +11,17 @@
     <artifactId>windup-reporting-impl</artifactId>
     <name>Windup Engine - Reporting Impl</name>
 
+    <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <excludes>
+                    <exclude>**/viz.js</exclude>
```
jsight: Why not just delete it?

jsight:
It looks like you have checked in a whole bunch of binary .class files as well somehow?

OndraZizka:
Binaries removed, no idea what created them.

reporting/api/src/main/java/org/jboss/windup/reporting/service/ReportService.java
OndraZizka: Ok. I can't get the context through @Inject, can I? So I am passing it from the RuleProvider.

reporting/impl/pom.xml
OndraZizka: Copy from where?
The thing with JS is that there's no mechanism to prevent JS being duplicated, i.e. some kind of resource dependencies. So if we had, say, viz.js multiple times, it could be in the code several times. Here it is in reporting impl, so kind of general storage for resources... no?

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/rendering/RenderGraphRuleProvider.java
```diff
@@ -17,13 +16,11 @@
...
 
 /**
- *
  * This renders the graph itself to the output directory for debug purposes.
  *
  * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
- *
  */
-@Vetoed
+@Vetoed // When enabling, also delete the <exclude>**/viz.js</exlude> from POM.
```
mareknovotny: why the exclusion should be deleted, when you removed the file below? That comment doesn't make sense to me.

OndraZizka: The file was duplicated in the code. Now it's there only once.

mareknovotny: ok, thanks for explanation.

reporting/impl/pom.xml
jsight: Yeah, an addon would be nice. We can worry about that later though.

PR #947, opened 2016-06-01
WINDUP-1059 Support hasHint for File condition

OndraZizka:
Needed by https://github.com/windup/windup-rulesets/pull/150

### DIFF
--- a/reporting/api/src/main/java/org/jboss/windup/reporting/config/HasHint.java
+++ b/reporting/api/src/main/java/org/jboss/windup/reporting/config/HasHint.java
@@ -10,6 +10,7 @@ import org.jboss.windup.graph.model.resource.FileModel;
 import org.jboss.windup.reporting.model.ClassificationModel;
 import org.jboss.windup.reporting.model.InlineHintModel;
 import org.jboss.windup.reporting.service.InlineHintService;
+import org.jboss.windup.rules.files.model.FileLocationModel;
 import org.jboss.windup.rules.files.model.FileReferenceModel;
 import org.jboss.windup.util.ExecutionStatistics;
 import org.ocpsoft.rewrite.context.EvaluationContext;
@@ -21,7 +22,7 @@ import org.ocpsoft.rewrite.param.RegexParameterizedPatternParser;
 /**
  * An implementation of {@link AbstractIterationFilter} to filter models based on the existence of a
  * {@link ClassificationModel} attached to the given payload.
- * 
+ *
  * @author <a href="mailto:lincolnbaxter@gmail.com">Lincoln Baxter, III</a>
  */
 public class HasHint extends AbstractIterationFilter<WindupVertexFrame> implements Parameterized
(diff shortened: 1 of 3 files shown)


OndraZizka:
Looks like I've changed the logic somehow.

reporting/api/src/main/java/org/jboss/windup/reporting/config/HasHint.java
```diff
@@ -34,51 +35,31 @@ public boolean evaluate(GraphRewrite event, EvaluationContext context, WindupVer
...
-                        if (parseResult.matches() && parseResult.isValid(event, context))
-                        {
-                            result = true;
-                            break;
-                        }
-                    }
-                }
-            }
+            Iterable<InlineHintModel> hints = null;
+            // FileLocationModel is a fake FileReferenceModel with [1,1] row and column, so we want hints for the whole file.
```
jsight: I do not understand this logic... I don't think that you can assume that a "FileReferenceModel" is a "fake".

OndraZizka: Yes, I've made a rushed conclusion. I think this will need a change in how `File` creates FileLocationModel.

OndraZizka: (16:56:00) ozizka-FN: jsightler:  See the PR ^
(16:56:19) ozizka-FN: Maybe I'm still following some bad logic. Looking at it
(16:56:38) ozizka-FN: Now I have to go pick up Hanka, but will look. Or tell me that it's really non-sense :)
(16:56:53) jsightler: Yeah, that logic is basically broken.
(16:59:18) ozizka-FN: But it seems to work, and also work for all tests, so far (still running).
(16:59:21) jsightler: Basically now if I actually do care about only hints that come file File.exists that will no longer be possible.
(16:59:40) jsightler: Its trading one set of limited semantics for a different set of limited semantics.
(17:00:05) ozizka-FN: jsightler:  Could you write a XML for that?
(17:00:11) jsightler: It would be far better to just make has-hint configurable.
(17:00:20) ozizka-FN: Hmm perhaps
(17:00:59) jsightler: (also, it would be better if rules like this were written in a rule language instead of the simplified xml syntax)
(17:01:01) ozizka-FN: <has-hint at="whole-file|line"> ?
(17:01:15) jsightler: Yeah, something like that.
(17:01:25) ozizka-FN: Well, that's completely different thing :)  Windup 4
(17:01:48) jsightler: Actually that isn't bad at all... that sort of allows it to generalize to cases where the condition isn't a File condition

OndraZizka: retest

OndraZizka:
Jess, you saying we should have the <has-hint at="whole-file|line">  in this PR?   If I understood well, the other thing you mentioned we would loose with this impl is just theoretical so far.

(16:56:53) jsightler: Yeah, that logic is basically broken.
(16:59:18) ozizka-FN: But it seems to work, and also work for all tests, so far (still running).
(16:59:21) jsightler: Basically now if I actually do care about only hints that come file File.exists that will no longer be possible.
(16:59:40) jsightler: Its trading one set of limited semantics for a different set of limited semantics.
(17:00:05) ozizka-FN: jsightler:  Could you write a XML for that?
(17:00:11) jsightler: It would be far better to just make has-hint configurable.
(17:01:01) ozizka-FN: <has-hint at="whole-file|line"> ?
(17:01:15) jsightler: Yeah, something like that.

d-s:
closing old stuff, rebase+reopen if still relevant

PR #949, opened 2016-06-02
    WINDUP-1043  Report index - 2 charts with duplicated information

OndraZizka:


OndraZizka:
I'll make it a bit narrower as shown during meeting.

OndraZizka:
Reorg

OndraZizka:
Reorg 2

OndraZizka:
Now the tables and charts are bound in one "cell", so they stack together on smaller devices.

mareknovotny:
fyi there is shown https://github.com/windup/windup/wiki/images/report-jee-example-application-report-index-page.png in https://github.com/windup/windup/wiki/Review-the-Report, which needs to replace new layout if this is merged

mareknovotny:
I like Reorg https://github.com/windup/windup/pull/949#issuecomment-223345800 , but there are tables too wide, so reducing table width would get more points from me.

d-s:
Let's set this aside for now. Tobias feels strongly about the presentation of this report, and I guess it was discussed a few months ago, so let's leave it as is for 2.6.

Maarc:
Nice changes :) In my custom reports I generate a view like "Reorg 2" or "Reorg" without tables. It makes it easier to explain.

jsight:
@Maarc I agree, although Tobias did not like it very much when we brought it up to him a few months ago. If we have agreement from you and him, I would be ok with merging this.

PR #951, opened 2016-06-03
Shared libs section in app list t145

OndraZizka:


graph/api/src/main/java/org/jboss/windup/graph/model/ProjectModel.java
```diff
@@ -101,18 +106,32 @@
     void setVersion(String version);
 
     /**
-     * Contains the project's name.
+     * Project's name. This is most often derived from project's root filename.
```
jsight: I don't think it is accurate that this is generally derived from the root filename. Most are derived from the maven name or manifest.mf name.

OndraZizka: Fixed.
     * Project's name.
     * This is often derived Maven name or MANIFEST.MF name or from project's root filename.

OndraZizka:
Rebased.

PR #953, opened 2016-06-03
WINDUP-1068: XML Schema Validation causes a couple of issues with eff…

jsight:
https://issues.jboss.org/browse/WINDUP-1068

Also fixes:
https://issues.jboss.org/browse/WINDUP-1070


OndraZizka:
Looks good to me too.

PR #956, opened 2016-06-07
WINDUP-1073: FileContent and File have a very slow approach to file search

jsight:
https://issues.jboss.org/browse/WINDUP-1073

### DIFF
--- a/graph/api/src/main/java/org/jboss/windup/graph/TitanUtil.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/TitanUtil.java
@@ -1,0 +1,18 @@
+package org.jboss.windup.graph;
+
+/**
+ * Contains methods that are useful for dealing with some of the peculiarities of the Titan data store and its indexing scheme.
+ *
+ * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
+ */
+public class TitanUtil
+{
+    /**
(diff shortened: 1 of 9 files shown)


mareknovotny:
indexing is a good approach, do we have a real numbers for comparison?

graph/api/src/main/java/org/jboss/windup/graph/TitanUtil.java
```diff
@@ -0,0 +1,18 @@
...
+public class TitanUtil
+{
+    /**
+     * This attempts to convert from standard regular expression syntax to a format that is suitable for
+     * a Lucene index search (such as those used by titan).
+     */
+    public static String titanifyRegex(String regex)
+    {
+        return regex.replace("\\Q", "\"").replace("\\E", "\"").replace("?:", "");
+    }
```
OndraZizka: This will only work for certain types of regexes?
Or any? I mean, will it change the meaning of the regex? How about
\QShe said "hello"\E
?

OndraZizka:
Good idea. Seems to work well.

graph/api/src/main/java/org/jboss/windup/graph/TitanUtil.java
jsight: Yes, it has all of the usual limitations of lucene indexes on regular expressions.

PR #958, opened 2016-06-08
Fixed a problem with the classification stats

jsight:
(this also adds a comment to the relevant code, as well as unit test so that we don't forget this edge case again)


d-s:
This fixes the regression I noticed between 06/03 and 06/07.

PR #960, opened 2016-06-09
WINDUP-1075: Duplicate archive tag attached, but duplicate list only …

jsight:
https://issues.jboss.org/browse/WINDUP-1075


d-s:
code + results lgtm

PR #961, opened 2016-06-09
Render the source report based upon a project parameter

jsight:
https://tree.taiga.io/project/rdruss-jboss-migration-windup-v3/task/167


jsight:
This is just for testing... the basic functionality works, but I think the source.ftl produces some slightly buggy html. Feel free to test and report bugs, though.

d-s:
So far unable to find any bugs. Haven't tested exhaustively but clicked into/out of source reports from migration issues, app details, remote services, EJBs, hard-coded IPs and haven't been switched into the wrong app from anywhere.

Not sure about what kind of issues you were looking for on the feedback form or how to look for them, but let me know tomorrow and I can QA more. I did notice that occasionally it gets shifted down about 20% and then the bottom 20% wraps up to the top, but I can't trigger that reliably to tell if it's new to this PR or not.

jsight:
Ok, I was hoping that the functionality works. The biggest issue remaining is that rendering the navbar multiple times results in multiple elements with the same id in the dom for a short period. There is a race condition here that could potentially cause issues and does cause some jquery issues in the console.

I think it will be easy to fix.

d-s:
@OndraZizka The results look fine to me, I think you can do better than me reviewing this code though, so please take a look.

reporting/impl/src/main/resources/reports/templates/include/navbar_macro.ftl
```diff
@@ -0,0 +1,52 @@
+<#macro renderNavbar applicationReportIndexModel index_page=false navUrlPrefix="">
+    <ul class="nav navbar-nav">
+    <#if applicationReportIndexModel ??>
```
mareknovotny: is this space before `??` typo ?

tests/src/test/java/org/jboss/windup/tests/application/WindupArchitectureDuplicateTest.java
```diff
@@ -79,7 +79,7 @@ public void testRunWindupDuplicateEAR() throws Exception
             validateMigrationIssues(context);
         } finally
         {
-            FileUtils.deleteDirectory(testTempPath.toFile());
+            //FileUtils.deleteDirectory(testTempPath.toFile());
```
mareknovotny: this probably needs to be uncommented while merging

jsight: Fixed

reporting/impl/src/main/resources/reports/templates/include/navbar_macro.ftl
jsight: Fixed

PR #963, opened 2016-06-10
Fix WINDUP-1081

OndraZizka:
https://issues.jboss.org/browse/WINDUP-1081


jsight:
At a glance, this looks good to me. I don't think that this code is currently used anywhere.

PR #964, opened 2016-06-10
Task 166: Only place a jar in shared archives if it is actually share…

jsight:
https://tree.taiga.io/project/rdruss-jboss-migration-windup-v3/task/166


d-s:
Everything looks good here.

PR #966, opened 2016-06-13
Taiga task#144 with some fixes

mareknovotny:
This uses commits from https://github.com/windup/windup/pull/961

Status of the tests is erratic/not passing as expected due below issues.

All reports have issues with navbars as there is NPE in reporModel.getReportPriority() and sorting of items blow up. As the opposite shared lib app reports don't suffer with this NPE. I can't explain why.


reporting/api/src/main/java/org/jboss/windup/reporting/model/JarDependenciesReportModel.java
```diff
@@ -0,0 +1,33 @@
...
+public interface JarDependenciesReportModel extends ApplicationReportModel
+{
+    String TYPE = "jarDependenciesReport";
+
+    String JAR_REPORT_TO_PROJECT_MODEL = "jarDepsToProjectModel"; 
+
+    /**
+     * Contains all {@link ProjectModel}s that contain this file
+     */
+    @Incidence(label = JAR_REPORT_TO_PROJECT_MODEL, direction = Direction.OUT)
```
jsight: This seems to create a link directly from the report to a JAR. I'm not sure that this will cover the relationship adequately as it seems that a report will have multiple groups of jars. It seems another vertex in between the report and the actual JAR might be useful.

If the words above are confusing, a picture might be clearer: https://docs.google.com/drawings/d/1mxGZN210tUEqie9fdqOMb41O3jz6HDyI8ZLNyhkTLJQ/edit

reporting/api/src/main/java/org/jboss/windup/reporting/model/JarDependencyReportToProjectEdgeModel.java
```diff
@@ -0,0 +1,41 @@
...
+import com.tinkerpop.frames.Property;
+import com.tinkerpop.frames.modules.typedgraph.TypeValue;
+
+
+/**
+ * @author mnovotny
+ *
+ */
+@TypeValue(JarDependencyReportToProjectEdgeModel.TYPE)
+public interface JarDependencyReportToProjectEdgeModel extends WindupEdgeFrame
```
jsight: Good, although I think that @InVertex will be something other than ProjectModel in your case.

tests/src/test/java/org/jboss/windup/tests/application/WindupArchitectureDuplicateTest.java
```diff
@@ -112,6 +115,16 @@ private void validateReportIndex(GraphContext graphContext)
         Assert.assertTrue(reportIndex.checkIncidentByCategoryRow("Potential Issues", 0, 0));
     }
 
+    private void validateJarDependencyReport(GraphContext graphContext)
```
jsight: Good idea to add a test here. :)

reporting/api/src/main/java/org/jboss/windup/reporting/model/JarDependenciesReportModel.java
mareknovotny: honestly I don't get what would be useful in this case. I thought that only thing I need is to get ProjectModel which holds the information about an archive and all duplicated archives which are already vertexes pointing to a Project model.

That is why I joined new Jar Dependency Report model to project model and archive model, where is the information about path.

jsight: The traversal can give you accurate path information based upon the application being traversed. It is basically impossible to get a full path back to root without one at this point.

I was thinking that there should be a link that looks something like this:

ReportModel -> Dependency (basically this groups archives by SHA1) -> Edge (with filename information attached to this edge) -> ArchiveModel

reporting/api/src/main/java/org/jboss/windup/reporting/model/JarDependencyReportToProjectEdgeModel.java
mareknovotny: if that is related to the comment https://github.com/windup/windup/pull/966#discussion_r66870655 I understand it like it should be connected with DependencyModel, right?

reporting/api/src/main/java/org/jboss/windup/reporting/model/JarDependenciesReportModel.java
mareknovotny: Hmm, so this is for first time I create a such thing, so be patient with me ;) as I need to get it correctly.

I understand it that we need to create new vertex, call it `DependencyModel` with sha1sum property and some properties to identify archive (model or just name) and create a java handler method to get all duplicate archives. Is that correct?
What I am not sure is about adding an edge, which is in this case `JarDependencyReportToProjectEdgeModel` which I will rename to more suitable name (for instance `JarDependencyReportToDependencyEdgeModel`) as I will replace projectModel there to point to a `ArchiveModel`.

reporting/api/src/main/java/org/jboss/windup/reporting/model/JarDependencyReportToProjectEdgeModel.java
jsight: Yes

reporting/api/src/main/java/org/jboss/windup/reporting/model/JarDependenciesReportModel.java
jsight: I do not think it will necessarily need a JavaHandler method. Using some specific names for the relationships, it might look something like this:

JarDependencyReportModel -> DependencyModel (SHA1:String) -> DependencyToArchiveEdgeModel (path:String) -> ArchiveModel

Feel free to use better names if you think of them.

It might also make sense to maintain a direct link from DependencyModel to the canonical project to make displaying the information on the report easier.

mareknovotny:
There are still minor issues with report template:
* sorting of file paths
* some information are hidden even I can see them in source of the report

I will fix it later tomorrow/ actually already today ;)

PR #968, opened 2016-06-13
WINDUP-1083: Remove dead doc link in Ignored Files report

d-s:
Removing the link entirely here rather than pointing it to somewhere else. Doesn't seem like our GH wiki is a stable enough place to point links at, and probably not worth the maintenance burden of keeping links from Windup output to RH Windup documentation in-sync unless we're going to do it in enough places to get significant value out of it.


windupgithubbot1:

windupgithubbot1:

OndraZizka:
test

PR #973, opened 2016-06-15
fixed all checks for empty element values in fm templates for pom.xml…

mareknovotny:
… file's


mareknovotny:
this is enhancement of https://github.com/windup/windup/pull/971f

PR #976, opened 2016-06-16
WINDUP-1092: Generating source reports is now spectacularly slow

jsight:
https://issues.jboss.org/browse/WINDUP-1092


d-s:
OK, I only ran one test with this (largest app with --packages com) but:
-- Wall time was about the same as before the spike
-- Same number of source reports as before the spike
-- Same incidents and story points (net of changes to rules in the interim)

I'd say:
-- Merge this today
-- Let the performance check run on everything overnight
-- Tomorrow morning, we'll check the performance reports, and I'll check a couple other apps to make sure there's no unexpected change to incident/point counts

PR #982, opened 2016-06-17
Task 159: Unify freemarker configuration to insure that localization …

jsight:
…and similar things would only need to be tweaked in one place

Relates to this, but this PR isn't really necessary as the default is sufficient:
https://tree.taiga.io/project/rdruss-jboss-migration-windup-v3/task/159


jsight:
retest

PR #983, opened 2016-06-20
taiga task #175 - show effort for shared-lib only if it is greater th…

mareknovotny:
…an 0

### DIFF
--- a/reporting/impl/src/main/resources/reports/templates/application_list.ftl
+++ b/reporting/impl/src/main/resources/reports/templates/application_list.ftl
@@ -38,12 +38,14 @@
                 <span class="points">${pointsFromAllTraversal}</span>
                 <span class="legend">story points</span>
             </div>
+            <#if virtualAppExists >
             <div class="effortPoints shared">
                 <#if appReport.projectModel.projectType! != "VIRTUAL">
                     <span class="points">${pointsFromSharedTraversal}</span>
                     <span class="legend">in shared libs <#--<br/>once: ${pointsFromOnceTraversal}--></span>
                 </#if>
             </div>
+            </#if>
             <div class="incidentsCount">
                 <table>
                     <#assign totalIncidents = 0 >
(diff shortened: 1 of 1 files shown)


reporting/impl/src/main/resources/reports/templates/application_list.ftl
```diff
@@ -38,12 +38,14 @@
                 <span class="points">${pointsFromAllTraversal}</span>
                 <span class="legend">story points</span>
             </div>
+            <#if (pointsFromSharedTraversal > 0) >
             <div class="effortPoints shared">
```
OndraZizka: I think this will make the elements right of this shift to the left.
I suggest to rather add display: hidden, in a way that you add a class pointsCount${pointsFromSharedTraversal?c} and
... .pointsCount0 { display: hidden; }

mareknovotny: yep, that is what I wanted, as keeping the space empty/not shifting looks weird IMO. I tried that approach

OndraZizka: BU wants this to keep a table-like style...

mareknovotny: and what breaks/change that?  it is stil in rows and columns rendered ;)

mareknovotny:
screenshot from 2016-06-20 16-01-01

OndraZizka:
Well the problem is, that sometimes some app has 0 and other has > 0, so the breakdown tables would be in different "columns".

d-s:
OK, sorry for the confusion, there are two separate cases to be addressed here:

(1) There are shared libraries, and we show the extra section for shared libraries, but some apps have 0 points from shared libraries. I agree with Ondra -- we should show a 0 for these libraries so we have a consistent display with the same number of columns in each row.

(2) There are no shared libraries and we do not display the extra section. For this case I say we should not show the shared/unique numbers -- they are not relevant since there are no shared libraries in this set of applications. In this case the display should look like what's in Marek's screenshot -- this doesn't cause any issues because every row will have the same columns displayed.

So to do this properly I think the #if (pointsFromSharedTraversal > 0) condition should be changed to a condition that checks whether we are displaying the shared libraries section.

@OndraZizka @mareknovotny Make sense?

mareknovotny:
@d-s yep.

mareknovotny:
So now after clarification, the shared lib points are rendered only if Shared library exists. If any application doesn't share anything it still shows the zero for comparison to others with some SP in shared lib app. Below is the screenshot:
screenshot from 2016-06-21 10-10-57

OndraZizka:
That's nice and usable. I already have a solution for that in PR #985 .
If you don't mind, I am closing this one.

mareknovotny:
@OndraZizka so if you are asking and close it immediately why are you asking?

PR #975, opened 2016-06-15
App list design T162

OndraZizka:
As per https://docs.google.com/document/d/1nnsmOu8INH_jxcyAYyry1TTtf47AMYE8BR9OjrGi3jI/edit?ts=575842d0

Each commit, starting from 81a94b2, is one version.
See https://drive.google.com/drive/u/1/folders/0B6TharS7fpqGLXBmNGhuaUpodXM for the pages.

PR #985, opened 2016-06-20
App list design3 feedback2

OndraZizka:
(I see no ID to refer to)
https://docs.google.com/document/d/1nnsmOu8INH_jxcyAYyry1TTtf47AMYE8BR9OjrGi3jI/edit?ts=575842d0

This is based on the #5 version from the other app list design PR #975 (== has shared commits)

2.6 App list tweaks based on discussion on 20160616:
Start with the latest version without the bubble (version #5 of the 6 Ondra shared)
Shared libs story point number feels a little too prominent
We need to communicate this number but it doesn't need to be so huge
Possible options:
 ✔ Replace the large shared-libs number with two smaller shared + unique numbers
Remove the extra column, and display the shared+unique numbers directly underneath the total story points number (possibly using the stacked-progress-bar style of visualization we talked about)
 ✔ There is quite a lot of horizontal whitespace if you are full-size on a large screen -- maybe the width of the row could be capped (say at ~1200px wide? or whatever looks good) (~1470 still looks good IMO)
 ✔ Incident count table: alignment is still a little bit different between app list rows if the largest number of digits is different (more obvious if you put in a 6 digit number) 
Minor issue, it looks perfect up to 4 digits

### DIFF
--- a/reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
+++ b/reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
@@ -78,7 +78,7 @@ public class CreateApplicationListReportRuleProvider extends AbstractRuleProvide
             {
                 applications.addItem(applicationReportModel);
                 if (ProjectService.SHARED_LIBS_UNIQUE_ID.equals(applicationReportModel.getProjectModel().getUniqueID()))
-                    relatedData.put("sharedLibsExists", applicationReportModel); // Instead of a boolean.
+                    relatedData.put("sharedLibsApplicationReport", applicationReportModel); // Used as kind of boolean in the template.
             }
         }
         relatedData.put("applications", applications);
--- a/reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
+++ b/reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
@@ -14,6 +14,7 @@ import org.jboss.windup.config.phase.PostReportGenerationPhase;
 import org.jboss.windup.graph.GraphContext;
 import org.jboss.windup.graph.model.WindupVertexFrame;
 import org.jboss.windup.graph.service.GraphService;
+import org.jboss.windup.graph.service.ProjectService;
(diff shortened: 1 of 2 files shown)


mareknovotny:
@OndraZizka could you share the screenshots with new renderings for all to get easier view on changes?

OndraZizka:
I've changed few bits, please review and merge if ok.

OndraZizka:
See https://drive.google.com/drive/u/1/folders/0B6TharS7fpqGLXBmNGhuaUpodXM - 11 and 12

d-s:
looks good to me.

jsight:
retest

reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
```diff
@@ -70,12 +71,16 @@ private void createIndexReport(GraphContext context)
...
         GraphService<WindupVertexListModel> listService = new GraphService<>(context, WindupVertexListModel.class);
         WindupVertexListModel<ApplicationReportModel> applications = listService.create();
+        Map<String, WindupVertexFrame> relatedData = new HashMap<>();
         for (ApplicationReportModel applicationReportModel : applicationReportService.findAll())
         {
             if (applicationReportModel.isMainApplicationReport() != null && applicationReportModel.isMainApplicationReport())
+            {
                 applications.addItem(applicationReportModel);
+                if (ProjectService.SHARED_LIBS_UNIQUE_ID.equals(applicationReportModel.getProjectModel().getUniqueID()))
+                    relatedData.put("sharedLibsExists", applicationReportModel); // Instead of a boolean.
```
jsight: Why not just use a boolean? The name is sort of odd for it not to be a boolean value.

OndraZizka: Because it's Map<String, WindupVertexModels> and I didn't want to put Object there... should I?

jsight:
retest

reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
OndraZizka: Changed the key name to sharedLibsApplicationReport

PR #986, opened 2016-06-20
App list sort T161

OndraZizka:


mareknovotny:
Before merging I would like to see these 3 commits squashed or fixed commit messages.

reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
```diff
@@ -69,15 +77,51 @@ private void createIndexReport(GraphContext context)
...
+            if (null == o2.getProjectModel() || null == o2.getProjectModel().getRootFileModel() || null == o2.getProjectModel().getRootFileModel().getFileName() )
+                return -1;
+
+            // On error resume next.
+            try {
+                return ObjectUtils.compare(
+                    o1.getProjectModel().getRootFileModel().getFileName(),
+                    o2.getProjectModel().getRootFileModel().getFileName());
+            }
+            catch (Throwable ex)
```
mareknovotny: Is this try-catch block necessary. I can see that all possible `Null` values are handled above and `ObjectUtils.compare` is according to javadoc of commons-lang3

> Null safe comparison of Comparables

mareknovotny:
I noticed check for Jenkins windup-pr-builder is not passing, exactly  org.jboss.windup.tests.application.WindupArchitectureDuplicateTest.testRunWindupDuplicateEAR test fails.

reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
```diff
@@ -69,15 +77,51 @@ private void createIndexReport(GraphContext context)
...
+        final Iterable<ApplicationReportModel> apps = applicationReportService.findAll();
+        List<ApplicationReportModel> appsList = new ArrayList();
+        for (ApplicationReportModel applicationReportModel : apps)
         {
             if (applicationReportModel.isMainApplicationReport() != null && applicationReportModel.isMainApplicationReport())
-                applications.addItem(applicationReportModel);
+                appsList.add(applicationReportModel);
         }
+        // Our current model doesn't keep the list order, but once I wrote, I'm leaving the sorting here for when it does.
+        appsList.sort(new AppRootFileNameComparator());
```
mareknovotny: This is Java 8 only method for `java.util.List`
Java 7 `java.util.List` doesn't have such method see https://docs.oracle.com/javase/8/docs/api/java/util/List.html#sort-java.util.Comparator- where it is described 

> since 1.8

jsight:
@mareknovotny They can be squashed in the github UI at merge time now.

jsight:
retest

mareknovotny:
yep, they can be, the only problem is if the message commit is not correctly describing what it changes, so if I see such commits I could squash it them by mistake, that is what i meant by the comment.

reporting/api/src/main/java/org/jboss/windup/reporting/rules/CreateApplicationListReportRuleProvider.java
OndraZizka: I changed this to be haltOnException = true, so I wanted to be super-defensive. If the sorting fails for whichever reason, it's better to show unsorted app list than no app list and fail :)

jsight:
Merged: https://github.com/windup/windup/commit/068fbb5479ac9103dd3ce7655b6cc1d1bf73f744

PR #990, opened 2016-06-21
Quick fix for a NPE, before a real fix.

OndraZizka:
See WINDUP-1097 for the issue.


d-s:
OK, this seems safe to me. I tested it, the rule stops failing, and the BOM has more entries since the rule didn't skip the rest of the IdentifiedArchives after the failing one. I think this is OK to merge.

jsight:
retest

OndraZizka:
Follow up in #991

PR #992, opened 2016-06-22
Fix intermittent failures with the unparsable test

jsight:


jsight:
Merging, as this fixes the tests on jenkins.

PR #993, opened 2016-06-23
Task177 and 178: Source and target version filtering

jsight:
This probably needs some discussion regarding the approach. The code seems to work, though.

Related PR: https://github.com/windup/windup-rulesets/pull/165


mareknovotny:
I briefly review it and it seems basically ok, I haven't tried it practically on a ruleset, but I will plan that.
Are you going to add an example of `*.windup.technologytransformer.xml` file into a test ?

jsight:
Yes, there will be a test and I am going to send a PR to rulesets with some
changes to support it as well.

OndraZizka:
The code looks good, but I am not quite sure what does it implement. The task describtions are very brief. What is the technologytransformer good for? Is there any overall description of that?

jsight:
@OndraZizka - The technologytransformer allows us to map from one technology id to another, along with version information.

This is primarily so that we can use version filtering instead of names without breaking user's expectations. So eap7 can be smoothly translated to eap:7 and things like that.

mareknovotny:
one additional thing in handling of CLI arguments is probably missing as I am testing similar launch command:
`bin/windup --input samples/jee-example-app-1.0.0.ear -target eap7`
and it returns validation error that such target doesn't exist
`ERROR: target value (eap7) not found, must be one of: [fuse, resteasy3, jbpm, eap, camel, linux, hibernate5, drools, fsw, hibernate-search5, hibernate, java-ee]`

mareknovotny:
so there is missing the changed validation of the option(s) for available targets/sources while transforming eap7 into eap:7

jsight:
Good catch... fixed with the latest commit.

mareknovotny:
great, from my POV, it is ready to be merged

OndraZizka:
Just an idea: I was thinking about adding version support also to tags. Then, this could be covered by the same functionality. The versions could be defined in tags definition files, as well as their relations. I.e.
```xml
<tag name="eap" version="7" role="technology">
      <tag name="eap7" role="technology"/>
```
But now that we have this... :)

jsight:
@OndraZizka I'm not sure that versions really make sense on tags. Although I do wish that we had some sort of metadata about the technologies themselves that was separate from the rules.

PR #997, opened 2016-07-05
Tweaks to the API to support windup-web

jsight:
Adds a convenience commit method and also makes the service() method return the interface that is proxyable instead of a concrete class.


jsight:
Merging to support further windup-web work.

PR #999, opened 2016-07-19
WINDUP-1108 TypeScript models generator

OndraZizka:
Creates TS models in ~/.windup/cache . Example:

```java
export class ClassificationModel
{
    private vertexId: number;

    private static const DISCRIMINATOR: string = 'ClassificationModel';

    ruleID: string;
    classification: string;
    description: string;

    function getAddFileModel(): FileModel {
        return this.queryAdjacent('ClassificationModel:classificationModelToFileModel', null);
    }

    function setAddFileModel(item: FileModel) {
        return this.set('ClassificationModel:classificationModelToFileModel', item);
    }


    function getFileModels(): FileModel {
        return this.queryAdjacent('ClassificationModel:classificationModelToFileModel', null);
    }

    function setFileModels(item: FileModel) {
        return this.set('ClassificationModel:classificationModelToFileModel', item);
    }


}
```

### DIFF
--- a/graph/impl/src/main/java/org/jboss/windup/graph/GraphContextImpl.java
+++ b/graph/impl/src/main/java/org/jboss/windup/graph/GraphContextImpl.java
@@ -1,5 +1,6 @@
 package org.jboss.windup.graph;
 
+import org.jboss.windup.graph.tsgen.TypeScriptModelsGenerator;
 import java.io.File;
 import java.lang.reflect.Method;
 import java.nio.file.Path;
@@ -25,7 +26,6 @@ import org.jboss.forge.furnace.util.Annotations;
 import org.jboss.windup.graph.frames.TypeAwareFramedGraphQuery;
 import org.jboss.windup.graph.listeners.AfterGraphInitializationListener;
 import org.jboss.windup.graph.listeners.BeforeGraphCloseListener;
-import org.jboss.windup.graph.model.WindupFrame;
 import org.jboss.windup.graph.model.WindupVertexFrame;
 import org.jboss.windup.graph.service.GraphService;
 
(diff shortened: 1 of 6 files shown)


mareknovotny:
what PR depends on?

mareknovotny:
just curious you know there are typescript generators could do it for you. Are they too universal or you like to implement the bridge yourself intentionally?

I see this is interesting from few others:
https://github.com/vojtechhabarta/typescript-generator
http://www.habarta.cz/typescript-generator/maven/typescript-generator-maven-plugin/generate-mojo.html

graph/impl/src/main/java/org/jboss/windup/graph/GraphContextImpl.java
```diff
@@ -85,6 +87,7 @@ public GraphContextImpl create()
         FileUtils.deleteQuietly(graphDir.toFile());
         TitanGraph titan = initializeTitanGraph();
         initializeTitanIndexes(titan);
+        generateTypeScriptModels();
```
mareknovotny: why are generated models everytime GraphContext is created? Aren't they the same until you change the code for Model (compile time) ?

OndraZizka: Yes, this is WIP and as the comment of the class says, it will be moved to a RuleProvider or even (better) to a standalone Windup command.

OndraZizka:
It depends on a PR that's was not created yet :) Now it's https://github.com/windup/windup-web/pull/19 and I also see Jess' https://github.com/windup/windup-web/pull/17 .

graph/impl/src/main/java/org/jboss/windup/graph/GraphContextImpl.java
jsight: Yes, this is the wrong place for the code, IMO. I don't think that is all that important right now. It really does need to just be a standalone tool eventually, though.

jsight:
IMO, this should be in a Maven plugin so that it can be executed as a build-time step by specifying an explicit goal.

jsight:
Isn't this kind of inventing an RPC mechanism for frames over REST? I guess that is one way to do it, though.

OndraZizka:
Continued (with changes) in https://github.com/windup/windup-web/pull/64

PR #1002, opened 2016-08-08
WINDUP-1099 adding resolving W3c dtd/xsd definitions locally and not …

mareknovotny:
…remotely

https://issues.jboss.org/browse/WINDUP-1099

This is a copy of https://github.com/mareknovotny/windup/commit/086c8434776b49f57931165c76ffba7907295c08 which goes into 2.7.x branch


rules-xml/api/src/main/resources/CatalogManager.properties
```diff
@@ -0,0 +1,8 @@
+#CatalogManager.properties
+verbosity=1
+relative-catalogs=false
+# Always use semicolons in this list
+catalogs=./core-catalog.xml
+prefer=public
+static-catalog=yes
+allow-oasis-xml-catalog-pi=yes
```
mareknovotny: This is a default settings for CatalogManager

mareknovotny: I added only W3C html/xhtml dtds/schemas. If we would like to optimize more validations like JBoss project/product dtds/schemas or purely Java EE ones, we can add later more definition in core-catalog.xml or add more catalogs for every possible thing.

jsight: Yes, I think we will need some Java EE ones as well, but this is a good start. +1

PR #1004, opened 2016-08-12
Fix graph key collision WINDUP-1120

OndraZizka:


jsight:
It looks like something is storing user data as property key names. That is probably a bad idea in general.

Anyway, this one is good to merge now.

PR #1008, opened 2016-08-23
added quickfix element for task 233

mareknovotny:
taiga task #231 Basic quick fix support in rules and tooling

### DIFF
--- a/graph/api/src/main/java/org/jboss/windup/graph/model/QuickfixModel.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/model/QuickfixModel.java
@@ -1,0 +1,55 @@
+package org.jboss.windup.graph.model;
+
+import com.tinkerpop.frames.Property;
+
+/**
+ * @author <a href="mailto:hotmana76@gmail.com">Marek Novotny</a>
+ *
+ */
+public interface QuickfixModel extends WindupVertexFrame
+{
+    String TYPE = "QuickfixModel";
+    String PROPERTY_TYPE = "type";
+    String PROPERTY_NAME = "name";
+    String PROPERTY_SEARCH_STRING = "search";
+    String PROPERTY_REPLACEMENT_STRING = "replacement";
+    String PROPERTY_NEWLINE_STRING = "newline";
+
+    /**
+     * Quickfix type {@link QuickfixType}
+     * @param type
+     */
+    @Property(TYPE)
+    void setQuickfixType(QuickfixType type);
+
+    /**
+     * Get the Quickfix type {@link QuickfixType}
+     * @return
+     */
+    @Property(TYPE)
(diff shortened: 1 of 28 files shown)


mareknovotny:
Implementation without updating/creating tests

graph/api/src/main/java/org/jboss/windup/graph/model/QuickfixModel.java
```diff
@@ -0,0 +1,55 @@
...
+    void setName(String name);
+
+    @Property(PROPERTY_NAME)
+    String getName();
+
+    @Property(PROPERTY_SEARCH_STRING)
+    void setSearch(String searchStr);
+
+    @Property(PROPERTY_SEARCH_STRING)
+    String getSearchStr();
```
jsight: Why is there a "Str" suffix on the getter methods?

tooling/api/src/main/java/org/jboss/windup/tooling/data/Classification.java
```diff
@@ -39,6 +39,12 @@
     Iterable<Link> getLinks();
 
     /**
+     * This contains a list of {@link Quickfix}s for follow up in tools
+     * @return
+     */
+    Iterable<Quickfix> getQuickfixes();
```
jsight: Do you think this will be applicable to Classifications as well?

jsight:
This looks good to me so far. I don't think the test needs to be particularly complicated. Just something that shows the handlers parse them correctly and that the quick fix gets put into the graph correctly should be sufficient.

I think you are really close at this point.

graph/api/src/main/java/org/jboss/windup/graph/model/QuickfixModel.java
mareknovotny: copy&paste error, I used that in parsing blocks somewhere else. Good catch!

tooling/api/src/main/java/org/jboss/windup/tooling/data/Classification.java
mareknovotny: Do you mean the multiple quickfixes or just quickfixes? I thought it is a rare case, but why not.

Actually I wasn't too convinced, but classification was in planning document.

mareknovotny:
Great, I will finish it today (your tomorrow).

tooling/api/src/main/java/org/jboss/windup/tooling/data/Classification.java
jsight: Oh, I had forgotten that this was in the planning document. I'm not entirely sure what the use case would be, but I guess there is no harm in leaving it here. Maybe that would be suitable for file level replacements (XSLT, etc).

jsight:
Great... merged!

PR #1009, opened 2016-08-26
WINDUP-1137: Auto-generated EAP XML descriptor files disappear betwee…

jsight:
https://issues.jboss.org/browse/WINDUP-1137


d-s:
Looks good, let's merge to 2.7.x branch as well since we'll want to include this if we do another release on that line.

PR #1010, opened 2016-08-30
Support custom rule paths that point to a file

jsight:


mareknovotny:
I guess we need to change also the '--userRulesDirectory' CLI option to omit Directory part and documentation as it is not only directory now, right?.

jsight:
@mareknovotny - Yes, that might be a good idea.

jsight:
JIRA for documentation: https://issues.jboss.org/browse/WINDUP-1141

jsight:
I am a little hesitant to change the name of an existing option. What do you think @d-s?

d-s:
Perhaps we could change the proper name to --userRulesPath and use that in 3.x documentation, and have a --userRulesDirectory option forwards to it for backwards-compatibility (like we did for the "eap7" technology)?

Doesn't necessarily need to be addressed right away in this PR either way.

mareknovotny:
@jsight you always could add new one to keep backward compatibility, that is the way it usually works the best. Not changing/removing but adding and deprecate it, but in this case deprecation is not our focus.

jsight:
@mareknovotny That is possible, although our option list is already long and ugly. I think we would need some way to hide the old one from the basic command line help if we are going to perpetually support old syntax.

mareknovotny:
@jsight true, some CLI programs has got basic help and help with advanced options.

jsight:
JIRA Filed: https://issues.jboss.org/browse/WINDUP-1144

PR #1024, opened 2016-10-12
Trimmed pom dependencies

jsight:
This brings the /lib folder from 85MB to 15MB.


mareknovotny:
that sounds great, what was the criteria to trim it?
I am asking as I can see many not used imports on classes in my IDE too, so maybe clearing them all could theoretically could go to next reducing the size.

jsight:
We only need public interfaces rather than implementations, so I tried to remove anything related to tinkerpop implementation.

I agree that we probably pull in things through the imports that are not actually used (or were used at one time and aren't now). There is probably another 10-20 MB that could be trimmed from that.

PR #1025, opened 2016-10-13
Cleaning of unused imports and useless overriden versions in pom.xml files

mareknovotny:


jsight:
I will wait on the tests to pass, but +1 otherwise. :) Thanks!

PR #1027, opened 2016-10-18
WINDUP-1196 Remove unused GetArchivesBySHA1Method which causes cyclic…

OndraZizka:
… deps


jsight:
I'm not sure how this could create a cyclic dependency? Reporting shouldn't depend on rules-java-archives?

jsight:
retest

PR #1028, opened 2016-10-20
WINDUP-1195 - --discoverPackages doesn't work correctly for .java source codes

klinki:
Class name should be based on class path relative from source root path.


windupgithubbot1:

windupgithubbot1:

mareknovotny:
@klinki Could you add a test for source input into [org.jboss.windup.tests.bootstrap.DiscoverPackagesCommandTest](https://github.com/windup/windup/blob/master/tests/src/test/java/org/jboss/windup/tests/bootstrap/DiscoverPackagesCommandTest.java) ?

mareknovotny:
retest

bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/DiscoverPackagesCommand.java
```diff
@@ -5,27 +5,12 @@
...
+import org.jboss.windup.bootstrap.commands.*;
 import org.jboss.windup.exec.configuration.options.InputPathOption;
 import org.jboss.windup.rules.apps.java.scan.operation.packagemapping.PackageNameMappingRegistry;
-import org.jboss.windup.util.ClassNameUtil;
-import org.jboss.windup.util.PackageComparator;
-import org.jboss.windup.util.PackageFrequencyTrie;
-import org.jboss.windup.util.PackageFrequencyTrieVisitor;
-import org.jboss.windup.util.PathUtil;
-import org.jboss.windup.util.ZipUtil;
+import org.jboss.windup.util.*;
```
jsight: +1

klinki:
I see your point. Question is, what do we do with that? I guess we cannot be sure about the structure of projects, right? They could use standard maven structure, but also some custom one. Or they could not use maven at all.

Can we really guess the package structure just by filepath? Or should we read content of file (in case of .java files it is easy) and read package structure from there?

jsight:
Yeah, I think they would have to be parsed for this to work. Maybe we can talk about whether this needs to be done now in the meeting? I don't think that we have ever really supported package filtering like this before, as it is fairly unusual to filter packages in source mode.

mareknovotny:
well, @klinki created the issue as he expected it should work for sources too, the problem here is that our CLI help says it should work for sources too. I would say we can go back and fix the help and documentation. I didn't know what is the correct behavior as documentation usually are taken as documented intention ;)

jsight:
I think that fixing the documentation should be sufficient for now. We can revisit package scanning of source code later if needed.

mareknovotny:
I sent https://github.com/windup/windup-documentation/pull/119 to documentation

mareknovotny:
Closing as we will do reimplementation later to have source folders correctly recognized

PR #1031, opened 2016-10-27
[WIP] - Separate freemarker

klinki:
In this PR I'm separating freemarker related code from reporting-api. I'm also modifying classes to be proxyable (adding empty protected constructor).


windupgithubbot1:

windupgithubbot1:

jsight:
I'm kind of +0 on this at the moment. Feel free to keep using it if it helps you keep moving forward, but it may turn out not to be necessary.

We can sort out how to proceed when we are a little further along (whether with this or something else or a mixture of things).

klinki:
I agree, with https://github.com/windup/windup-web/pull/75 this is probably not necessary anymore. Closing for now. Maybe we will get back to it sometimes later.

PR #1035, opened 2016-11-04
Story 302: Severity category work

jsight:
Not quite ready to review yet from a code standpoint, but it should be feature complete now.


jsight:
retest

jsight:
Related: https://github.com/windup/windup-rulesets/pull/177

reporting/api/src/main/java/org/jboss/windup/reporting/severity/IssueCategoryRegistry.java
```diff
@@ -0,0 +1,170 @@
...
+
+    /**
+     * Adds a {@link IssueCategory} to the registry and throws a {@link DuplicateIssueCategoryException} if the item already exists.
+     */
+    public void addCategory(IssueCategory category) throws DuplicateIssueCategoryException
+    {
+        IssueCategory original = this.issueCategories.get(category.getCategoryID());
+
+        // Just overwrite it if it was a placeholder
+        if (!original.isPlaceholder())
```
d-s: I think you get NPE here once you start adding categories to your XML registry that aren't in the defaults you're adding as placeholders?

jsight: Nice catch... fixed (and test-case added). :)

jsight:
@d-s - The pom isn't including the issue descriptions, and that is causing the order to be wrong on the report index. I will look into it for the application list as well.

reporting/api/src/main/java/org/jboss/windup/reporting/config/classification/Classification.java
```diff
@@ -49,20 +51,10 @@
...
-    {
-        return quickfixes;
-    }
-
     private RegexParameterizedPatternParser classificationPattern;
     private RegexParameterizedPatternParser descriptionPattern;
     private int effort;
-    private Severity severity = EffortReportModel.DEFAULT_SEVERITY;
-
+    private IssueCategory issueCategory = null;
```
mareknovotny: what about default severity/category ? Some hints/classification uses default setting. Are there migrated to explicit values in the following PR in windup-rulesets?

reporting/api/src/main/java/org/jboss/windup/reporting/severity/DuplicateIssueCategoryException.java
```diff
@@ -0,0 +1,16 @@
+package org.jboss.windup.reporting.severity;
```
mareknovotny: is *severity* now category, so why we have package as `org.jboss.windup.reporting.severity` and not `org.jboss.windup.reporting.category`?

config-xml/schema/windup-jboss-ruleset.xsd
```diff
@@ -282,19 +282,15 @@
                 <xs:element ref="quickfix" minOccurs="0" maxOccurs="unbounded"/>
             </xs:sequence>
             <xs:attribute name="title" type="xs:string" use="required"/>
-            <xs:attribute name="severity" type="severity-type" use="optional" />
+            <xs:attribute name="categoryID" type="category-id-type" use="optional" />
```
jsight: I know that is true for element names, but I wasn't sure if we wanted the same convention for attribute ids. Do you think "category-id" would be better?

reporting/api/src/main/java/org/jboss/windup/reporting/config/classification/Classification.java
jsight: It is still setting the category to a default, it is just doing it farther down in the code now.

reporting/api/src/main/java/org/jboss/windup/reporting/severity/DuplicateIssueCategoryException.java
jsight: Good question. I will change it to category.

config-xml/schema/windup-jboss-ruleset.xsd
mareknovotny: We uses already capital case like afterMethod. So i think going with lowercase or capitalcase is what we have - categoryId or category-id -> choose what you think is better ;)

reporting/api/src/main/java/org/jboss/windup/reporting/config/classification/Classification.java
mareknovotny: ok, good, I see it now in IssueCategoryRegistry ;)

jsight:
@mareknovotny - Changes made as requested... I think this is ready to review again.

mareknovotny:
great, it looks fine

jsight:
@mareknovotny @johnsteele - Will it cause any issues with the Eclipse tooling to merge it now?

It will require some changes if you are currently tracking master, but probably not if you are on 3.0.0.Alpha1.

PR #1038, opened 2016-11-21
Added serialization support to the tooling API

jsight:


jsight:
retest

jsight:
retest

tooling/impl/src/main/java/org/jboss/windup/tooling/ExecutionResultsImpl.java
```diff
@@ -33,37 +41,67 @@
  * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
  * @author <a href="mailto:hotmana76@gmail.com">Marek Novotny</a>
  */
+@XmlRootElement(name = "execution-results")
 public class ExecutionResultsImpl implements ExecutionResults
 {
```
OndraZizka: Isn't this class mixing data and logic too much? Maybe methods working with GraphContext could be somewhere else or at least static?

OndraZizka:
Leaving to John to decide

tooling/impl/src/main/java/org/jboss/windup/tooling/ExecutionResultsImpl.java
jsight: Yes, that is probably true. It could probably use some refactoring.

PR #1040, opened 2016-11-22
WINDUP-1239 NPE in DiscoverDuplicateArchiveProjectsRuleProvider

OndraZizka:


jsight:
In what scenario does this happen?

jsight:
Good catch... it looks like this happens because the jar in a source project may not necessarily have a parent jar. There are actually two jars in the built version that are identical, and neither are "sub" jars, so this can happen.

I think this fix will work fine.

jsight:
Thanks... also the diagram at the top seems useful. :)

PR #1042, opened 2016-11-24
updated to Furnace 2.24.3.Final

mareknovotny:


mareknovotny:
reference fix in the update https://issues.jboss.org/browse/FURNACE-127

PR #1043, opened 2016-11-26
3. Tech report server resources - task 335

OndraZizka:


jsight:
I guess this doesn't have the elements for discovering these items yet? Are you planning to add that next?

OndraZizka:
Yes, there already is the task for datasources.

PR #1044, opened 2016-11-29
T379 Make it possible to stop running Windup core execution

OndraZizka:
This makes use of WindupProgressMonitor#isCancelled().
That is queried in DefaultRuleLifecycleListener
which is queried by RuleSubset. 
RuleSubset, on spoting stop request, stops the rules cycle and creates a WindupExecutionModel vertex.

### DIFF
--- a/config/api/src/main/java/org/jboss/windup/config/AbstractRuleLifecycleListener.java
+++ b/config/api/src/main/java/org/jboss/windup/config/AbstractRuleLifecycleListener.java
@@ -6,26 +6,29 @@ import org.ocpsoft.rewrite.context.EvaluationContext;
 /**
  * This provides a set of default (empty) methods that make it easy to implement {@link RuleLifecycleListener}s that only need to override a subset of
  * the available methods.
- * 
- * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
  *
+ * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
+ * @author <a href="mailto:zizka@seznam.cz">Ondrej Zizka</a>
  */
 public abstract class AbstractRuleLifecycleListener implements RuleLifecycleListener
 {
-
     @Override
     public void beforeExecution(GraphRewrite event)
     {
     }
 
     @Override
-    public void beforeRuleEvaluation(GraphRewrite event, Rule rule, EvaluationContext context)
+    public boolean beforeRuleEvaluation(GraphRewrite event, Rule rule, EvaluationContext context)
     {
+        // Execution was not cancelled.
+        return false;
     }
 
     @Override
-    public void ruleEvaluationProgress(GraphRewrite event, String name, int currentPosition, int total, int timeRemainingInSeconds)
+    public boolean ruleEvaluationProgress(GraphRewrite event, String name, int currentPosition, int total, int timeRemainingInSeconds)
     {
+        // Execution was not cancelled.
+        return false;
     }
 
     @Override
@@ -34,8 +37,10 @@ public abstract class AbstractRuleLifecycleListener implements RuleLifecycleList
     }
 
     @Override
-    public void beforeRuleOperationsPerformed(GraphRewrite event, EvaluationContext context, Rule rule)
+    public boolean beforeRuleOperationsPerformed(GraphRewrite event, EvaluationContext context, Rule rule)
     {
+        // Execution was not cancelled.
+        return false;
     }
 
     @Override
--- a/config/api/src/main/java/org/jboss/windup/config/GraphRewrite.java
+++ b/config/api/src/main/java/org/jboss/windup/config/GraphRewrite.java
@@ -50,6 +50,7 @@ public class GraphRewrite extends AbstractRewrite implements Rewrite
 
     /**
      * Stores the exception which holds information if, and where, the Windup stopped (typically on an external request).
+     * If windup was stopped, this must be called.
      */
     public void setWindupStopException(WindupStopException windupStopException)
     {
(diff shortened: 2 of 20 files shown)


config/api/src/main/java/org/jboss/windup/config/AbstractRuleLifecycleListener.java
```diff
@@ -6,26 +6,29 @@
...
 {
-
     @Override
     public void beforeExecution(GraphRewrite event)
     {
     }
 
     @Override
-    public void beforeRuleEvaluation(GraphRewrite event, Rule rule, EvaluationContext context)
+    public boolean beforeRuleEvaluation(GraphRewrite event, Rule rule, EvaluationContext context)
```
jsight: I think that it would be simpler to add a "isCancelled" flag rather than have every method return a boolean. It feels like it is overloading the functionality of these listener methods to add this as a return value here.

config/api/src/main/java/org/jboss/windup/config/GraphRewrite.java
```diff
@@ -34,12 +39,34 @@ public GraphRewrite(Iterable<RuleLifecycleListener> listeners, GraphContext cont
...
+    {
+        return windupStopException;
+    }
+
+    /**
+     * Stores the exception which holds information if, and where, the Windup stopped (typically on an external request).
+     */
+    public void setWindupStopException(WindupStopException windupStopException)
+    {
+        if (this.windupStopException != null)
```
jsight: Why not just create a new WindupStopException with the existing one as the "cause" here?

config/api/src/main/java/org/jboss/windup/config/operation/Iteration.java
```diff
@@ -284,9 +285,13 @@ else if (condition != null)
                     variables.pop();
                 }
             }
+            catch (WindupStopException ex)
+            {
+                throw new WindupStopException("Windup stop requested in " + this.toString(), ex);
```
jsight: Can't you just rethrow it here?

graph/api/src/main/java/org/jboss/windup/graph/model/WindupExecutionModel.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.tinkerpop.frames.Property;
+import com.tinkerpop.frames.modules.typedgraph.TypeValue;
+
+/**
+ * Information about the Windup execution.
+ * Some of the information are stored in memory, see e.g. {@link RuleExecutionResultsListener}.
+ */
+@TypeValue(WindupExecutionModel.TYPE)
+public interface WindupExecutionModel extends WindupVertexFrame
```
jsight: Good idea :)

config/api/src/main/java/org/jboss/windup/config/GraphRewrite.java
OndraZizka: Because second call to this method would mean there's a broken logic somewhere in the stop handling code.

config/api/src/main/java/org/jboss/windup/config/AbstractRuleLifecycleListener.java
OndraZizka: I think this approach is simpler. Not necessarily more intuitive, but it's keeping the API minimal and reducing the logic. Plus, it's done. No business value in changing it now :) But if you like, I can change.

OndraZizka:
org.jboss.windup.tests.bootstrap.ListTechnologiesAndTagsCommandsTest.tags
Stacktrace
java.lang.AssertionError
	at org.jboss.windup.tests.bootstrap.ListTechnologiesAndTagsCommandsTest.tags(ListTechnologiesAndTagsCommandsTest.java:27)

Standard Output
> JBoss Windup, version 3.0.0-SNAPSHOT. JBoss Forge, version 2.24.3.Final
Available tags:

Standard Error
Failed to load rule information due to: null

OndraZizka:
Not sure how the failure relates to this PR

OndraZizka:
Not sure how the failure relates to this PR

OndraZizka:
retest

config/api/src/main/java/org/jboss/windup/config/operation/Iteration.java
jsight: Good point, +1

jsight:
np

config/api/src/main/java/org/jboss/windup/config/AbstractRuleLifecycleListener.java
jsight: Ok, lets keep it this way for now.

config/api/src/main/java/org/jboss/windup/config/GraphRewrite.java
```diff
@@ -34,12 +39,35 @@ public GraphRewrite(Iterable<RuleLifecycleListener> listeners, GraphContext cont
...
+    public WindupStopException getWindupStopException()
+    {
+        return windupStopException;
+    }
+
+    /**
+     * Stores the exception which holds information if, and where, the Windup stopped (typically on an external request).
+     * If windup was stopped, this must be called.
+     */
+    public void setWindupStopException(WindupStopException windupStopException)
```
jsight: I'm still not sure how calling this twice would necessarily be wrong. Couldn't that just mean that the user had clicked "cancel" more than once or something?

OndraZizka: I'll have a look at that again. But how I intended it when I had it freshly in mind, having another call would imply there is a broken logic in the caller. Let me check

jsight: Oh, I see that now. Thanks for clarifying. :)

PR #1046, opened 2016-12-13
Added different method for tag comparison.

klinki:
I had issue with very non intuitive tag include/exclude in windup-web.

I added method which returns true if:
- tag is included or tag is not excluded

It checks if includeTags is not empty, if it is not, tag is included ONLY when it is in includeTags.
If includeTags is empty, tag is by default included, but is excluded, when it is in excludeTags.


windupgithubbot1:

windupgithubbot1:

jsight:
@windupgithubbot1 Test this please

PR #1049, opened 2016-12-16
adding timestamps to reports

mareknovotny:
This is easier to recognize if reports are from the last or from old processing.


jsight:
Nice... thanks!

PR #1050, opened 2016-12-16
WINDUP-1265 - Added ability to terminate Windup while running in serv…

johnsteele:
…er mode from the ExecutionBuilder.


windupgithubbot1:

PR #1051, opened 2016-12-19
Tech report filter

klinki:
Preparing common interface for filtering models by project.

### DIFF
--- a/graph/api/src/main/java/org/jboss/windup/graph/model/resource/FileModel.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/model/resource/FileModel.java
@@ -3,6 +3,9 @@ package org.jboss.windup.graph.model.resource;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.EnumUtils;
@@ -222,6 +225,10 @@ public interface FileModel extends ResourceModel, BelongsToProject
     @JavaHandler
     boolean belongsToProject(ProjectModel projectModel);
 
+    @Override
+    @JavaHandler
+    Iterable<ProjectModel> getRootProjectModels();
+
     abstract class Impl implements FileModel, JavaHandlerContext<Vertex>, BelongsToProject
     {
         public ProjectModel getApplication()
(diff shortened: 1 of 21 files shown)


windupgithubbot1:

windupgithubbot1:

mareknovotny:
retest

graph/api/src/main/java/org/jboss/windup/graph/model/resource/FileModel.java
```diff
@@ -322,5 +327,11 @@ public File asFile() throws RuntimeException
 
             return new File(getFilePath());
         }
+
+        @Override
+        public boolean belongsToProject(ProjectModel projectModel)
+        {
+            return this.getProjectModel().equals(projectModel);
```
jsight: You might want to verify that the project being passed in is a canonical project first. Check if it is an instance of DuplicateProjectModel and if it is, then use the result of calling getCanonicalProject() on it.

graph/api/src/main/java/org/jboss/windup/graph/service/ProjectService.java
```diff
@@ -67,4 +78,62 @@ public ProjectModel getOrCreateSharedLibsProject()
 
         return sharedLibsProject;
     }
+
+    public Map<ProjectModel, ProjectModel> getProjectToRootProjectMap()
```
jsight: I think that this code needs javadoc. It isn't totally clear to me what it is doing or why it should be used vs just calling getRootProject() on a given project.

graph/api/src/main/java/org/jboss/windup/graph/service/ProjectService.java
```diff
@@ -67,4 +78,62 @@ public ProjectModel getOrCreateSharedLibsProject()
...
+
+        for (FileModel inputPath : WindupConfigurationService.getConfigurationModel(this.getGraphContext()).getInputPaths())
+        {
+            ProjectModel rootProjectModel = inputPath.getProjectModel();
+            if (rootProjectModel == null)
+            {
+                continue;
+            }
+
+            ProjectModelTraversal traversal = new ProjectModelTraversal(rootProjectModel, new OnlyOnceTraversalStrategy());
```
jsight: This should probably be the All strategy rather than "OnlyOnce", if I have read the intent correctly here.

jsight:
I know that this is WIP, but I added a few comments. Hopefully they are helpful. :)

graph/api/src/main/java/org/jboss/windup/graph/service/ProjectService.java
klinki: I'm not sure, this is based on this code: https://github.com/windup/windup-web/blob/master/addons/web-support/impl/src/main/java/org/jboss/windup/web/addons/websupport/rest/MigrationIssuesEndpointImpl.java#L72

from MigrationIssues.

klinki:
Thanks, I will have a look.

graph/api/src/main/java/org/jboss/windup/graph/service/ProjectService.java
klinki: Ah, right. I missed this one, but I think you are right. Calling getRootProject should be enough.

klinki:
retest

jsight:
retest

jsight:
Merged (and yeah, I should have squashed first... oops).

PR #1056, opened 2017-01-03
Fix attempt

jsight:


OndraZizka:
Better than what's in master now - merging

PR #1058, opened 2017-01-04
Class cast fix

jsight:
Experiment...

### DIFF
--- a/graph/api/src/main/java/org/jboss/windup/graph/GraphTypeManager.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/GraphTypeManager.java
@@ -122,7 +122,7 @@ public class GraphTypeManager implements TypeResolver, FrameInitializer
             return;
         String typeValue = typeValueAnnotation.value();
 
-        AbstractElement abstractElement = GraphTypeManager.asTitanVertex(element);
+        AbstractElement abstractElement = GraphTypeManager.asTitanElement(element);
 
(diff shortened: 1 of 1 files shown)


graph/api/src/main/java/org/jboss/windup/graph/GraphTypeManager.java
```diff
@@ -270,21 +270,21 @@ public static boolean hasType(Class<? extends WindupVertexFrame> type, Vertex v)
...
     {
         if (e instanceof StandardVertex)
         {
             return (StandardVertex) e;
         }
         else if (e instanceof EventVertex)
         {
-            return (StandardVertex) ((EventVertex) e).getBaseVertex();
+            return (AbstractElement) ((EventVertex) e).getBaseVertex();
         } else if (e instanceof StandardEdge)
```
OndraZizka: Note that the original code casted to Titan's StandardVertex too. Is this solving some CCEx?

OndraZizka: Jess said, yes, but for some reason it didn't get hit before...

jsight: Just to clarify, it isn't because I see it happening now, it is simply no longer necessary to cast to StandardVertex here.

jsight:
Thanks @mareknovotny for noticing the class cast error that was the root cause here.

PR #1072, opened 2017-02-03
WINDUP-1298: Really don't make network calls in offline

jsight:
This is still a little weird, as this creates some validation errors on poms and things like that. I really wonder if we should just skip this rule entirely in offline.

### DIFF
--- a/rules-xml/impl/src/main/java/org/jboss/windup/rules/apps/xml/xml/EnhancedEntityResolver2.java
+++ b/rules-xml/impl/src/main/java/org/jboss/windup/rules/apps/xml/xml/EnhancedEntityResolver2.java
@@ -13,9 +13,9 @@ import org.xml.sax.ext.EntityResolver2;
 
 /**
  * This is an {@link EntityResolver2} that has been enhanced to support URL redirection.
- * 
+ *
  * It also use local stored DTDs/XSDs as possible not to slow down validation.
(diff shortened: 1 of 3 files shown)


OndraZizka:
Provided it doesn't fail the run, I think it's ok.
Maybe we could bundle more XSDs? E.g. for the pom

jsight:
Yeah, I think bundling maven xsds would make sense.

d-s:
Yeah, seems like we will get "XML File is not valid" hits on every app now. How was it making calls after the "return null;" before?

jsight:
The xerces resolver would call this one and if it returned null, then it would try to resolve it itself if it looked like a URL. I think it treats "null" as an indication that a resolver doesn't know how to resolve a particular type rather than as an indication of failure.

I'm going from reading the implementation, though, rather than from reading docs as I haven't seen clear documentation on this.

jsight:
It looks like this will break a few tests because of increasing the number of failures.

d-s:
Don't really love spitting out clsf/hint everywhere that XML files are not valid when we didn't try to validate them. Kind of prefer your suggestion above, that we just not run this rule if we're offline.

OndraZizka:
So maybe catch the exception somewhere up and just print a log message?
INFO if offline, WARN if online

mareknovotny:
I think this should not be returning null it just give a chance to others resolvers  in chain.

We need to skip the rule completely.

d-s:
looks good here

PR #1073, opened 2017-02-05
WINDUP-1303 Provide DTOs for sharing Rule Provider info through tooli…

johnsteele:
…ng API.


jsight:
@johnsteele  - Sent a PR here to extend the functionality for this:
https://github.com/johnsteele/windup/pull/2

Origin should work as well. See the unit test for an example.

PR #1074, opened 2017-02-06
Remove effort from WebLogic EJB XML clsf that has a hint with effort.

d-s:
Set effort to 0 in clsf generated by ResolveWebLogicEjbXmlRuleProvider to avoid duplicating effort from weblogic-xml-descriptors-04000 hint.

Hint with effort was added after 2.7 in this commit: https://github.com/windup/windup-rulesets/commit/baf299554f86d834c9e9188d9d06477cc3ed08b6


d-s:
I tested this on my side and it checks out. Really we shouldn't have 2 rules generating the same classification, or a mandatory classification with 0 effort, but that seems like a fix for a day-that's-not-release-day.

jsight:
lgtm

PR #1079, opened 2017-03-13
WINDUP-1325 Fix tests errors

juangon:
This is my first PR as Red Hat employee :-).

Just notice me if you see anything wrong.

Thanks!

### DIFF
--- a/java-ast/tests/pom.xml
+++ b/java-ast/tests/pom.xml
@@ -41,6 +41,12 @@
             <artifactId>windup-exec</artifactId>
             <classifier>forge-addon</classifier>
             <scope>test</scope>
+        </dependency>    
+        <dependency>
+            <groupId>org.eclipse.tycho</groupId>
+            <artifactId>org.eclipse.jdt.core</artifactId>
+            <version>3.10.0.v20140604-1726</version>
+            <scope>test</scope>
         </dependency>
 
     </dependencies>
--- a/test-util/src/main/java/org/jboss/windup/testutil/html/TestJavaApplicationOverviewUtil.java
+++ b/test-util/src/main/java/org/jboss/windup/testutil/html/TestJavaApplicationOverviewUtil.java
@@ -77,6 +77,7 @@ public class TestJavaApplicationOverviewUtil extends TestReportUtil
                                 ".//div[contains(@class, 'container') and contains(@class, 'mainGraphContainer')]//div[@class = 'points']/div[@class = 'number']"));
         String effortString = effortElement.getText().trim();
         effortString = effortString.replace(",", "");
+        effortString = effortString.replace(".", "");
 
         int effort;
         try


windupgithubbot1:

windupgithubbot1:

mareknovotny:
retest

test-util/src/main/java/org/jboss/windup/testutil/html/TestJavaApplicationOverviewUtil.java
```diff
@@ -77,6 +77,7 @@ public void checkMainEffort(int expectedEffort)
                                 ".//div[contains(@class, 'container') and contains(@class, 'mainGraphContainer')]//div[@class = 'points']/div[@class = 'number']"));
         String effortString = effortElement.getText().trim();
         effortString = effortString.replace(",", "");
+        effortString = effortString.replace(".", "");
```
mareknovotny: This seems to me like you have set up different locale environment than UK/US for number delimiters. We have on Jenkins CI for master branch and this failure doesn't show up there.

java-ast/tests/pom.xml
```diff
@@ -41,6 +41,12 @@
             <artifactId>windup-exec</artifactId>
             <classifier>forge-addon</classifier>
             <scope>test</scope>
+        </dependency>    
+        <dependency>
+            <groupId>org.eclipse.tycho</groupId>
+            <artifactId>org.eclipse.jdt.core</artifactId>
+            <version>3.10.0.v20140604-1726</version>
+            <scope>test</scope>
```
mareknovotny: This should be unnecessary dependency as there is provided `org.jboss.windup.ast:windup-java-ast:forge-addon` dependency

juangon: Ok @mareknovotny, but I just didn't do anything. Just clone, and mvn clean install. AFAIK there is no mention to any required locale in order those tests to pass. Am I wrong?

juangon: BTW @mareknovotny my OS locale is in spanish (LANG=es_ES.UTF-8). Didn't know this can have impact in tests. Additionally, I don't think this is related to the first test failure. Can you try testing in a empty environment? (maybe remove all in your local .m2 repo and clone windup master again).

mareknovotny: @juangon our CI tests does that on clean environment and didn't fail on that I am sure. But I can give a try on my isolated env too, just for sure.

Wrt locale, if I use my native czech locale i could see the test failures too in past, there is testing of a number in thousands so there are dots/commas in formatted efforts in reports depending what locale can freemarker use.

juangon: Thanks @mareknovotny. Do you know which Maven version is CI using? In my case it's 3.3.9.

mareknovotny: Maven 3.2.5 but that should not be a cause, I am switching between 3.3.9 and 3.2.5 regularly

test-util/src/main/java/org/jboss/windup/testutil/html/TestJavaApplicationOverviewUtil.java
jsight: It might be useful if our tests did pass in other locales, though. Maybe this would be sufficient?

java-ast/tests/pom.xml
jsight: +1 - This dependency shouldn't be necessary.

juangon:
Resent here https://github.com/windup/windup/pull/1082 with the supposed right fix.

PR #1085, opened 2017-03-16
Registry approach to custom quickfix classes

jsight:
This is the framework discussed here:
https://docs.google.com/document/d/1k1DTcXg1YZk0dJVB8ceIJLGYtTiZu2cStzPw50-HW28/edit

@johnsteele - Not tested really, but this is the basic concept. Is this enough to get you going with your addon?


johnsteele:
@jsight Yes, this is excellent. Thanks!

jsight:
BTW, I think your rule might actually be possible as a pure xml rule. You
may only need the Java code for the quickfix itself.

jsight:
Closing (integrated into John's branch already).

PR #1091, opened 2017-03-28
Migration Issues filtering

klinki:
Create new ProblemSummaryServiceForWindupWeb - it groups results based on ProjectModel - for easy filtering on frontend.

### DIFF
--- a/reporting/api/src/main/java/org/jboss/windup/reporting/freemarker/problemsummary/ProblemSummaryServiceForWindupWeb.java
+++ b/reporting/api/src/main/java/org/jboss/windup/reporting/freemarker/problemsummary/ProblemSummaryServiceForWindupWeb.java
@@ -1,0 +1,220 @@
+package org.jboss.windup.reporting.freemarker.problemsummary;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


jsight:
I think that it would be better if this were in windup-web rather than here. It should work fine from the web support addon there.

klinki:
I guess you are right..
But I was thinking about creating rule for it as we have for Tech. report. (https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/TechnologiesStatsRuleProvider.java)

So we could compute it once, store results in graph and then only read it and not compute again. What do you think?

jsight:
That is fine, although if the rule is executed only for windup-web, the windup-web related add-on is still a better place for it. :)

jsight:
I agree that precomputing it and storing it in the graph would be nice.

mareknovotny:
@klinki so if you change the class name to have it more like general service not specifically bound to Windup Web usage, I am fine to have it in windup core too otherwise I agree with @jsight the proper place is windup web support addon.

klinki:
Sure, this is definitely not a final class name :)  I just didn't know how to name it properly. You know, naming, the most difficult thing in programming :)

PR #1096, opened 2017-03-29
WINDUP-1320 and tasks 608,607 - Report index charts and tables relayout

mareknovotny:
* I changed generation of issues summaries to have categoryID instead of name of category
* i change layout to be more like dynamic report index with hiding/showing tables 

I need to fix mandatory chart as it is somehow crippled with incidents and if it could be done x axis labels to be rotated and possibly fit longer categoryIds - that is one option.

### DIFF
--- a/reporting/impl/src/main/resources/reports/resources/css/windup.css
+++ b/reporting/impl/src/main/resources/reports/resources/css/windup.css
@@ -369,3 +369,8 @@ strong.wu-navbar-header {
     text-transform: uppercase;
     font-size: 18px;
 }
+
+.chartBoundary h4 {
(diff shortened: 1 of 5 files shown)


mareknovotny:
There is still possibility to have more categories with non-zero incidents in charts and then it will overflow each other. 
The question is if in that case should we change the size of chart to be wider or cut the x axis labels. The latter could lead to nonsenses I guess.

mareknovotny:
Need to fix tests with display hidden table values

mareknovotny:
ref. JIRA issue https://issues.jboss.org/browse/WINDUP-1320

PR #1099, opened 2017-04-05
Fix charts to use category id and not display 0 items

jsight:
Just until something like https://github.com/windup/windup/pull/1096 can be approved.


mareknovotny:
I don't understand if you took https://github.com/windup/windup/pull/1096 and changed a little it or you still are waiting on merging https://github.com/windup/windup/pull/1096. 
I see changes in master branch, but not all things like css styles.

jsight:
Extensive changes like this to this page require a review from Tobias and Marc. So I did a much smaller version.

We can present this to them when we have time, though.

mareknovotny:
:o there were just syncing to have the similar what is in dynamic report index

jsight:
I know, but those are completely new and are being reviewed weekly (or at least they should be). This could come as an unexpected surprise.

PR #1100, opened 2017-04-07
Console progress output on the same line

mrizzi:
1. pad the n-th message to be long as the (n-1)-th message to have every message clearly written without the final part of the previous one
2. print the n-th message with the "\r" carriage return to stay on the same line
3. println the next rule execution message with "\r" prefix to have the log clear with just all the rules executed


windupgithubbot1:

PR #1107, opened 2017-05-03
T896 Hardcoded IP cloud-mandatory

OndraZizka:
I am not exactly sure about the changes here. Because I don't exactly understand the existence of cloud-mandatory category. Feels like some confusion of terms.

### DIFF
--- a/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverHardcodedIPAddressRuleProvider.java
+++ b/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverHardcodedIPAddressRuleProvider.java
@@ -20,6 +20,7 @@ import org.jboss.windup.graph.GraphContext;
 import org.jboss.windup.graph.model.FileLocationModel;
 import org.jboss.windup.graph.model.resource.SourceFileModel;
 import org.jboss.windup.graph.service.GraphService;
+import org.jboss.windup.reporting.category.IssueCategoryRegistry;
 import org.jboss.windup.reporting.model.ClassificationModel;
 import org.jboss.windup.reporting.service.ClassificationService;
 import org.jboss.windup.rules.apps.java.model.PropertiesModel;
@@ -38,7 +39,7 @@ import org.w3c.dom.Element;
  * @author <a href="mailto:bradsdavis@gmail.com">Brad Davis</a>
  * @author <a href="mailto:hotmana76@gmail.com">Marek Novotny</a>
  */
-@RuleMetadata(phase = MigrationRulesPhase.class)
+@RuleMetadata(phase = MigrationRulesPhase.class, tags = {"cloud-readiness"})
 public class DiscoverHardcodedIPAddressRuleProvider extends AbstractRuleProvider
 {
     private static final String IP_PATTERN = "(?<![\\w.])\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}\\.\\d{1,3}(?![\\w.])";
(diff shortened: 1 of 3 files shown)


rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverHardcodedIPAddressRuleProvider.java
```diff
@@ -87,7 +91,8 @@ public void perform(GraphRewrite event, EvaluationContext context, FileLocationM
                     hintBody.append("\n\n");
                     hintBody.append("When migrating environments, hard-coded IP addresses may need to be modified or eliminated.");
                     location.setHint(hintBody.toString());
-
+                    //location.setIssueCategory(IssueCategoryRegistry.loadFromGraph(context, IssueCategoryRegistry.MANDATORY));
+                    location.setIssueCategory(IssueCategoryRegistry.loadFromGraph(context, "cloud-mandatory");
```
jsight: Please add an entry to the issuecategoryregistry instead of just using a string literal here.

jsight:
Just one small thing, otherwise this looks good.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverHardcodedIPAddressRuleProvider.java
mareknovotny: But we talked about to have it as an addition, right? Static IP addresses detection is important for other targets too even you are not doing cloud-readiness target.

jsight: Yes. I don't think this changes the target, though, so it should still fire even with non "cloud-readiness" migrations.

mareknovotny: well, true, category-id.

jsight: I am a little concerned that the reports will look strange, but I guess that is what was asked for.

mareknovotny:
Compilation error:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:compile (default-compile) on project windup-rules-java-api: Compilation failure
[ERROR] /var/lib/jenkins/workspace/windup-pr-builder/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverHardcodedIPAddressRuleProvider.java:[95,109] error: ')' expected
[ERROR] -> [Help 1]

OndraZizka:
Compilation works fine for me

OndraZizka:
Constant added. Compilation - I fixed the call and forgot to push afterwards. Fixed ;-)

mareknovotny:
I guess you need to check expected SP too as you raised the effort about 1 for every occurrence. Not sure if there is only 3 times that occurrence though.

See test failure:

> Effort was 2213 but was expected to be 2210

mareknovotny:
@OndraZizka there is also a request to change the title - removing word **detected** at
https://github.com/OndraZizka/windup/blob/eb1fa5761b224cc1471203e951a28dfc702c773f/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/ip/DiscoverHardcodedIPAddressRuleProvider.java#L80

PR #1110, opened 2017-05-05
T874 Add black brand bar at the top - T875Remove main logo

mrizzi:
- create new component `navheader.ftl` 
- add style for components of the brand bar (`class="wu-navbar-header"`)
- change all the pages to included the `navheader.ftl` component with the brand bar
- add link to `navbar.js` in pages where it was missing to be sure that `div id="main-navbar"` is rightly positioned (`about_windup.ftl, ruleprovidersummary.ftl, windupfreemarkerfunctions.ftl`)
- remove main logo (task#875)
- fixed a closing tag in https://github.com/windup/windup/blob/master/reporting/impl/src/main/resources/reports/templates/classloader.ftl#L78

### DIFF
--- a/reporting/impl/src/main/resources/reports/templates/include/navheader.ftl
+++ b/reporting/impl/src/main/resources/reports/templates/include/navheader.ftl
@@ -1,0 +1,9 @@
+            <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-responsive-collapse">
+                <span class="icon-bar"></span>
+                <span class="icon-bar"></span>
+                <span class="icon-bar"></span>
+            </button>
+            <span class="wu-navbar-header">
+              <strong class="wu-navbar-header">Red Hat Application Migration Toolkit</strong>
+               - Web Console 
+            </span>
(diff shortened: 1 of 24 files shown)


windupgithubbot1:

mareknovotny:
@mrizzi it seems there is a suspicious merge commit and it anyway is required to do a rebase. Have you had some issues with rebase?

reporting/impl/src/main/resources/reports/templates/include/navheader.ftl
```diff
@@ -0,0 +1,9 @@
+            <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-responsive-collapse">
+                <span class="icon-bar"></span>
+                <span class="icon-bar"></span>
+                <span class="icon-bar"></span>
+            </button>
+            <span class="wu-navbar-header">
+              <strong class="wu-navbar-header">Red Hat Application Migration Toolkit</strong>
+               - Web Console
```
mareknovotny: Web Console? this is RHAMT CLI, right? Even we share that in Web UI, this is IMHO wrong to be here.

mrizzi:
@mareknovotny it happened that when i did the PR, github told me that my PR was not merge-able for a conflict and i solved it immediately through github web UI

reporting/impl/src/main/resources/reports/templates/include/navheader.ftl
mrizzi: Yep, you're right, the report comes from CLI even if it's opened from Web UI.
So, the official name to put in the bar is "CLI" instead of "Web Console", right?

mareknovotny:
merged with rebase as https://github.com/windup/windup/commit/67f6ed688

PR #1111, opened 2017-05-09
T921 Support *.rhamt.xml files

mrizzi:
- `org.jboss.windup.config.parser.XMLRuleProviderLoader` added `rhamt.xml` extension handling
- `org.jboss.windup.util.furnace.FurnaceClasspathScanner.scanForAddonMap(Predicate<String>)` now adds an `Addon` to the result `Map` [only if the `discoveredURLs` is not empty](https://github.com/windup/windup/compare/master...mrizzi:taiga_T921?expand=1#diff-3a906af129780ff3ec760dc872a80ba8R67). In the case of running it two times ([one for each extension](https://github.com/windup/windup/compare/master...mrizzi:taiga_T921?expand=1#diff-b5e0dd11f5f1abceb2ed9530c72e5d99R158)), this change is necessary otherwise the second time the `Addon` with the first extension will be added with an empty `discoveredURLs`. 
- `org.jboss.windup.tooling.rules.RuleProviderRegistryImpl.getProviderType(String)` with its test `org.jboss.windup.tooling.ExecutionBuilderTest.testRuleProviderRegistry()`
- `org.jboss.windup.config.parser.XMLRuleProviderLoaderTest`[#77](https://github.com/windup/windup/compare/master...mrizzi:taiga_T921?expand=1#diff-9a9ec48ea0a43161570c6c0318933bccR77): get by `indexOf` because `org.jboss.windup.config.parser.XMLRuleProviderLoader.getProviders(RuleLoaderContext)` cycles through a `Map` implemented by an `java.util.IdentityHashMap<K,V>` that "[makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.](https://docs.oracle.com/javase/8/docs/api/java/util/IdentityHashMap.html)"
- updated many tests that uses a `*.windup.xml` file to also have a test for a `*.rhamt.xml` file (e.g. `org.jboss.windup.reporting.handlers.*`)

_NOT CHANGED_
- `org.jboss.windup.tests.bootstrap.AbstractBootstrapTestWithRules` otherwise all the tests that extend it should be duplicated

_TO DISCUSS_
- Is `org.jboss.windup.config.parser.XMLRuleProviderLoaderTest` with the two deployments the right way to change the tests with `AddonArchive` and `Deployments`?

### DIFF
--- a/config-xml/addon/src/main/java/org/jboss/windup/config/parser/XMLRuleProviderLoader.java
+++ b/config-xml/addon/src/main/java/org/jboss/windup/config/parser/XMLRuleProviderLoader.java
@@ -203,7 +203,9 @@ public class XMLRuleProviderLoader implements RuleProviderLoader
         }
     }
 
-    private boolean pathMatchesNamePattern(Path file) {
-        return file.getFileName().toString().toLowerCase().endsWith("." + XML_RULES_WINDUP_EXTENSION) || file.getFileName().toString().toLowerCase().endsWith("." + XML_RULES_RHAMT_EXTENSION);
+    private boolean pathMatchesNamePattern(Path file)
+    {
+        return file.getFileName().toString().toLowerCase().endsWith("." + XML_RULES_WINDUP_EXTENSION)
+                    || file.getFileName().toString().toLowerCase().endsWith("." + XML_RULES_RHAMT_EXTENSION);
     }
 }
--- a/config-xml/addon/src/main/java/org/jboss/windup/config/parser/XMLRuleProviderLoader.java
+++ b/config-xml/addon/src/main/java/org/jboss/windup/config/parser/XMLRuleProviderLoader.java
@@ -40,7 +40,7 @@ import org.w3c.dom.Document;
 
 /**
  * This {@link RuleProviderLoader} searches for and loads {@link AbstractRuleProvider}s from XML files that within all
- * addons, with filenames that end in ".windup.xml".
+ * addons, with filenames that end in ".windup.xml" or ".rhamt.xml".
  *
  * @author <a href="mailto:lincolnbaxter@gmail.com">Lincoln Baxter, III</a>
  *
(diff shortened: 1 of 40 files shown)


windupgithubbot1:

config-xml/addon/src/main/java/org/jboss/windup/config/parser/XMLRuleProviderLoader.java
```diff
@@ -201,6 +204,6 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
     }
 
     private boolean pathMatchesNamePattern(Path file) {
-        return file.getFileName().toString().toLowerCase().endsWith("." + XML_RULES_EXTENSION);
+        return file.getFileName().toString().toLowerCase().endsWith("." + XML_RULES_WINDUP_EXTENSION) || file.getFileName().toString().toLowerCase().endsWith("." + XML_RULES_RHAMT_EXTENSION);
```
mareknovotny: This is not readable much, can you format according to our format style? We have 150 columns on line at max.

config-xml/tests/src/test/java/org/jboss/windup/config/parser/XMLRuleProviderLoaderTest.java
```diff
@@ -44,11 +46,19 @@
     })
     public static AddonArchive getDeployment()
     {
-        return ShrinkWrap.create(AddonArchive.class)
+        return  ShrinkWrap.create(AddonArchive.class)
                     .addBeansXML()
```
mareknovotny: Why special space there ?

mareknovotny:
retest

mareknovotny:
retest

jsight:
I think this is fine. I don't believe that it is necessary to have separate deployments here:
https://github.com/mrizzi/windup/blob/2bb0e036f774c1ce24dbf6ca3f151f24d4493686/config-xml/tests/src/test/java/org/jboss/windup/config/parser/XMLRuleProviderLoaderTest.java#L54

That could just be part of the first deployment, IMO. Otherwise, I don't see any issues.

+1 - ready to merge from my standpoint

PR #1112, opened 2017-05-10
T876 remove license labels from apps list

OndraZizka:
Marc said this task is only about removing DiscoverArchiveLicenseFilesRuleProvider.java which is not complete.
I am putting this here because it has some things I did before I started improving how the tags are created.


jsight:
I think this is obsolete now?

PR #1115, opened 2017-05-15
T842 Add the path to the archive that failed to unzip

OndraZizka:
The task wanted a link. But is that the right solution? Is the user expected to download it and try to unzip manually? Although, we would have to store the archive exceptionally as a part of the report, which we normally don't do unless --keepWorkDirs .

So I am only adding the path.

### DIFF
--- a/graph/api/src/main/java/org/jboss/windup/graph/model/ProjectModel.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/model/ProjectModel.java
@@ -52,6 +52,10 @@ public interface ProjectModel extends WindupVertexFrame
     @Adjacency(label = ROOT_FILE_MODEL, direction = Direction.OUT)
     void setRootFileModel(FileModel fileModel);
 
+    /**
+     * This represents the root directory (in the case of a source-based analysis) or root archive (for binary analysis) containing this particular
+     * project.
+     */
     @Adjacency(label = ROOT_FILE_MODEL, direction = Direction.OUT)
     FileModel getRootFileModel();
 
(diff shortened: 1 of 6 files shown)


OndraZizka:
I have tried to test this but it seems the unparsable.jar in our sample .ear isn't processed.

OndraZizka:
The file is being skipped entirely, needs debugging.

mareknovotny:
retest

mareknovotny:
retest - after fixing windup-rulesets master

mareknovotny:
retest

jsight:
I think that it should only show the path instead of adding a link.

OndraZizka:
Now I looked at it closer and the thing is that the file name doesn't appear. This PR makes the issue have the file name but I guess it wouldn't appear in the report. However I can't make the invalid zipfile appear at all.

OndraZizka:
@jsight  Where are ClassificationModels converted to HintModels? At least that's my assumption, see my comment at https://tree.taiga.io/project/rdruss-jboss-migration-windup-v3/task/842?kanban-status=543761

OndraZizka:
Ok never mind, I read on and found that it's iterating over clsf too...

OndraZizka:
So the bug goes down to 
```
{l:"${renderedLink?json_string}", oc:"${fileSummary.occurrences?json_string}"},
```
looking at what exactly is wrong

OndraZizka:
Ok now it should be fixed properly. Works for me.

mareknovotny:
@OndraZizka  Do you  have somewhere test case or what application is producing the output for Malformed zip file?

PR #1117, opened 2017-05-26
tasks 1009/1002/1007 - disabling classification due duplications

mareknovotny:
…l-descriptor-04000 XML rule

https://tree.taiga.io/project/rdruss-jboss-migration-windup-v3/task/1009


mareknovotny:
it seems the failed test in windup-rulesets relies on java rule which is adding the classification. I removed the test as that should be testing only result of xml rules.

mareknovotny:
retest

jsight:
+1

PR #1118, opened 2017-05-26
task 998 - removing filename of Malformed zip file from classificatio…

mareknovotny:
…n title

https://tree.taiga.io/project/rdruss-jboss-migration-windup-v3/task/998


jsight:
+1

PR #1123, opened 2017-06-07
WINDUP-1431: rhamt-cli should not allow a target of eap without a ver…

jsight:
…sion

closes WINDUP-1431

### DIFF
--- a/bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
@@ -222,8 +222,7 @@ public class Bootstrap
 
     public static String getVersionString()
     {
-        return "> JBoss Windup, version " + getRuntimeAPIVersion() + ". JBoss Forge, version "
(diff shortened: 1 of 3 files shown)


bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
```diff
@@ -222,8 +222,7 @@ public static String getVersion()
 
     public static String getVersionString()
     {
-        return "> JBoss Windup, version " + getRuntimeAPIVersion() + ". JBoss Forge, version "
-                    + AddonRepositoryImpl.getRuntimeAPIVersion();
+        return "> Red Hat Application Migration Toolkit (RHAMT) CLI, version " + getRuntimeAPIVersion() + ".";
```
mareknovotny: I thought that is not part of rebranding ;), good catch!

jsight: I'm really surprised that noone else has pointed it out.

PR #1126, opened 2017-06-13
Revert "task 998 - removing filename of Malformed zip file from class…

OndraZizka:
…ification title"

This reverts commit 7f2ba3167b565c432f109a7a2f8cca451b06b8d3.


OndraZizka:
So the agreed way of fixing this is to have a list of files in the single one classification.

PR #1127, opened 2017-06-13
Use FQCN for logger

klinki:
- logger must use fully qualified class name (to support filtering by "org.jboss.windup")


jsight:
+1 - I noticed a couple of these while looking at the analysis logs in the other PR. Feel free to merge it whenever you feel that it is complete.

PR #1130, opened 2017-06-16
WINDUPRULE-172 Change default mode to DETAIL_ONLY.

OndraZizka:
As discussed on the meeting:
For users, this default will make more sense. For us, it will mean having extra issue-display-mode=all, because we misuse <classification> as a file-level hint (and that should be fixed).
Also see https://github.com/windup/windup-rulesets/pull/222


mareknovotny:
I think this confuses some tests as they are trying to find concrete issues.

OndraZizka:
This change would require to fix many test rules, so I am postponing it.

jsight:
+1 - There is no hurry with this, IMO.

PR #1141, opened 2017-07-12
WINDUP-1548 Show "DOES THIS STILL HAPPEN?" only once (from Parameteri…

OndraZizka:
…zedGraphCondition)

I  have set successive occurences to Level.FINER and added some more useful log message.


jsight:
lgtm! It will be interesting to see if performance #s improve any with the nightly runs.

PR #1143, opened 2017-07-13
Windup 1548 fix

Maarc:
Avoid building the very costly to build String if there is no need to log it.


OndraZizka:
I forgot to make it a lambda...

jsight:
Fixed in a slightly different way by @OndraZizka here: https://github.com/windup/windup/pull/1144

PR #1148, opened 2017-07-27
WINDUP-1572: replaced Windup strings in user facing messages

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1572


jsight:
lgtm - This should probably be merged to both branches.

mareknovotny:
yes, this is more related to 4.0 than everything else ;)

PR #1149, opened 2017-07-29
WINDUP-1595 Javassist is out of date in Tattletale

OndraZizka:
Let's try this. I guess it needs building and running with that app.
I don't know which app that is though. Couldn't reproduce.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -111,6 +111,12 @@
                 <artifactId>json</artifactId>
                 <version>20160810</version>
             </dependency>
+            <!-- WINDUP-1595  The version in Tattletale's POM is obsolete 3.15, let's force newer. -->
+            <dependency>
+                <groupId>org.javassist</groupId>
+                <artifactId>javassist</artifactId>
+                <version>3.21.0-GA</version>
+            </dependency>        
         </dependencies>
     </dependencyManagement>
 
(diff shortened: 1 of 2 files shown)


mareknovotny:
Why 3.18.0-GA was chosen? 
I looked at release notes where it says full support of java 8 and I found only this comment at https://github.com/jboss-javassist/javassist/releases/tag/rel_3_19_0_ga

So we should upgrade to that or to https://github.com/jboss-javassist/javassist/releases/tag/rel_3_20_0_ga ?

mareknovotny:
I have additional commit with managing dependencies even for other usage of javassist https://github.com/mareknovotny/windup/commit/8a78de9e67dca10d7c6283b5f27900896ffdda82

OndraZizka:
I relied on the maven index which shown me 3.18 as the newest.

OndraZizka:
Right, that's a good idea to have it centrally... maybe :) Assuming Javassist is fully backward compatible

OndraZizka:
So let's use 3.21? Do you have a PR?

OndraZizka:
I have moved the version declaration to <dependencyManagement> in root POM.

mareknovotny:
there is still windup-frames hardcoded version, but I can fix it while merging

PR #1154, opened 2017-08-21
WINDUPRULE-276: new Cache rule providers

mareknovotny:
https://issues.jboss.org/browse/WINDUPRULE-276

Some rule providers are checking the file names, but these seem to me a little weak pattern.


mareknovotny:
Apart wrong regex pattern, The new rules are triggered on weird occurences of "cache" pattern. Discovering

jsight:
+1 to merging this to 4.0.x and master. I have similar concerns about the rules, but I don't think that is a blocker unless you have found something specific.

Feel free to merge it whenever you'd like. :)

PR #1155, opened 2017-08-21
Switched parsing to a different batch approach to improve performance…

jsight:
… with large numbers of input applications


jsight:
For the moment, this PR is just here to get tests running. I ran some tests yesterday with a huge batch of input applications and ran into a couple of surprising performance issues with decompilation and Java parsing. The Java parser (JDT) seems to spend an inordinate amount of time in classpath checking, and maybe reducing the library path to only relevant jars can improve that.

jsight:
retest

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/model/HasApplications.java
```diff
@@ -19,29 +19,10 @@
...
-    boolean belongsToProject(ProjectModel projectModel);
-
-    @Override
-    @JavaHandler
     Iterable<ProjectModel> getRootProjectModels();
 
     abstract class Impl implements HasApplications, JavaHandlerContext<Vertex>, BelongsToProject
     {
-        @Override
-        public boolean belongsToProject(ProjectModel project)
```
jsight: This change is unrelated, but I removed it as the operation was unused and inherently unsafe. It doesn't deal with shared libraries correctly.

PR #1159, opened 2017-08-22
WINDUP-1620: text changes

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1620

all commits after review should be squashed, but for now i have kept it as it was in JIRA listed in that order.

### DIFF
--- a/reporting/impl/src/main/resources/reports/templates/application_list.ftl
+++ b/reporting/impl/src/main/resources/reports/templates/application_list.ftl
@@ -168,16 +168,12 @@
                     <div class="main">Application List</div>
                 </h1>
                 <div class="desc">
-                    The Application List report shows all applications which were analyzed.
-                    Click on an individual application to see individual reports or you can follow
-                    the global Migration Issues report.
+                    This report lists all analyzed applications. Click on an individual application to show more details.
                 </div>
             </div>
         </div>
 
-
         <!-- Apps -->
-
         <#assign sharedLibsExists = reportModel.relatedResources["sharedLibsApplicationReport"]!?has_content >
 
         <section class="apps">
--- a/reporting/impl/src/main/resources/reports/templates/application_list.ftl
+++ b/reporting/impl/src/main/resources/reports/templates/application_list.ftl
@@ -218,9 +218,9 @@
         </#if>
 
         <div style="width: 100%; text-align: center">
-            <a href="reports/windup_ruleproviders.html">Executed rules overview</a>
+            <a href="reports/windup_ruleproviders.html">Rule providers execution overview</a>
                 |
-            <a href="reports/windup_freemarkerfunctions.html">${getWindupBrandName()} FreeMarker methods</a>
+            <a href="reports/windup_freemarkerfunctions.html">FreeMarker methods</a>
         </div>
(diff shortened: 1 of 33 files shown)


reporting/impl/src/main/resources/reports/templates/application_list.ftl
```diff
@@ -223,9 +218,9 @@
         </#if>
 
         <div style="width: 100%; text-align: center">
-            <a href="reports/windup_ruleproviders.html">Executed rules overview</a>
+            <a href="reports/windup_ruleproviders.html">Rule providers execution overview</a>
```
mareknovotny: This doesn't include the title change, not sure why, but if that should be consistent I will change that.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateReportIndexRuleProvider.java
```diff
@@ -27,7 +27,7 @@
 @RuleMetadata(phase = ReportGenerationPhase.class)
 public class CreateReportIndexRuleProvider extends AbstractRuleProvider
 {
-    public static final String REPORT_INDEX = "Report Index";
+    public static final String REPORT_INDEX = "Dashboard";
```
mareknovotny: Not sure why this "Dashboard" is better title.

jsight: Well, if he removes the report list, then it is no longer an index to reports. I can see his point here.

jsight:
lgtm -- Is there an easy way to merge this to both branches?

mareknovotny:
I am not aware easy way to do it from github UI.

On terminal I do simple cherry-pick of the commit or continuous sequence of commits (commit1^..commitN), this is safer way in my opinion how to do it as the parent doesn't have to be the same on both branches. And to avoid anything in between, it is better to squash into one commit the original PR too.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateReportIndexRuleProvider.java
mareknovotny: What is the Report list? Do you mean removal of Additional Reports from https://github.com/windup/windup/pull/1158 ?

jsight: Yes, I meant the additional report links at the bottom of the page. We used to treat this page as a guide to the other reports rather than as a "dashboard", but I think that Marc views it differently.

jsight:
@mareknovotny - Oh, that is what I have been doing as well. :) I guess that is the best way, but it has the downside of not having a clear history in the github UI (its hard to tell which PRs are merged to which branches when looking at old PRs).

mareknovotny:
yep, if there is stable and development branch the PR workflow should be stricter and with using only baseline branch for all PR even it is more work, but we are not there with our side 4.0.x branch which should be temporary, but is living too long now

PR #1161, opened 2017-08-23
WINDUP-1619: Marc's style changes

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1619

I will try to fix the changes differently not including the table row in layout, but if you want to push this PR I don't care, but I think we can do better and simpler.


jsight:
I will go ahead and push it. I can't say that I like the way the dependency report looks now, though. :(

mareknovotny:
@jsight yeah, it is not consistent with others.

- panels styles are used wrongly and there is **h4** headings instead of **h3** and it seems results as an additional space before each dependency library item. We need to fix it.

PR #1162, opened 2017-08-24
Changed description to not confuse later for such description even id…

mareknovotny:
…entified

I saw multiple times in Dependency report where all fields are filled and Description with "Unidentified Archive" string.


jsight:
Good idea.

PR #1163, opened 2017-08-24
WINDUP-1625: Switched parsing to a different batch approach to improv…

jsight:
…e performance with large numbers of input applications


graph/api/src/main/java/org/jboss/windup/graph/model/BelongsToProject.java
```diff
@@ -5,13 +5,6 @@
  */
 public interface BelongsToProject
 {
-    /**
```
jsight: This change is unrelated, but I removed it as the operation was unused and inherently unsafe. It doesn't deal with shared libraries correctly.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/decompiler/FernflowerDecompilerOperation.java
```diff
@@ -314,7 +314,7 @@ private void handleDecompiledJavaFile(FileModel decompiledFileModel, FileService
             queueSize.incrementAndGet();
             executorService.submit(saveDecompiledRunnable);
 
-            while (queueSize.get() > 1000)
+            while (queueSize.get() > 100000)
```
mareknovotny: why 100K ? Is that calculated or from experienced testing?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/decompiler/FernflowerDecompilerOperation.java
```diff
@@ -326,6 +326,9 @@ private void handleDecompiledJavaFile(FileModel decompiledFileModel, FileService
                     // noop
                 }
             }
+
+            if (queueSize.get() > 1000)
+                LOG.warning("Decompiler queue size is: " + queueSize.get());
```
mareknovotny: what is this warning good for? when we have another for 100K above? I understand that first warning but the second is redundant or not?

graph/api/src/main/java/org/jboss/windup/graph/model/BelongsToProject.java
klinki: It is actually not unused, it is used here: https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/model/stats/TechnologiesStatsService.java#L259

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/decompiler/FernflowerDecompilerOperation.java
jsight: Yes, I was finding that this queue was getting exceptionally large on some of the large batchs. I think this was largely the result of analyzing obfuscated code, but I'm not sure that I see a harm to a large queue size here.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/decompiler/FernflowerDecompilerOperation.java
jsight: It shouldn't be there. I used that to monitor the queue size while testing other options than 5k. I will remove this line.

jsight:
Merging -- master only (4.1) as we haven't really tested this thoroughly enough for a 4.0 release, IMO.

graph/api/src/main/java/org/jboss/windup/graph/model/BelongsToProject.java
mareknovotny: I get the previous comment as "It is actually used,..." :)

But the referenced  TechnologyStatsService usage is not using the removed method.

I find it https://github.com/windup/windup-web/blob/master/addons/web-support/impl/src/main/java/org/jboss/windup/web/addons/websupport/rest/graph/EJBResourceImpl.java#L68 which now show the failure in http://wonka.mw.lab.eng.bos.redhat.com/jenkins/job/windup-web-nightly/687/

PR #1167, opened 2017-09-05
WINDUP-1630: Technology stats data model update

jsight:


rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyUsageStatisticsService.java
```diff
@@ -0,0 +1,384 @@
...
+            }
+        }
+
+        if (result == null)
+        {
+            result = create();
+            result.setComputed(new Date());
+            result.setProjectModel(projectModel);
+            result.setName(technologyName);
+            result.setOccurrenceCount(0);
```
klinki: Why is default value `1` if you always set it to `0`?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -0,0 +1,146 @@
...
+    {
+        this.technologyName = technologyName;
+    }
+
+    /**
+     * Creates the operation with the given input variable and technology name.
+     */
+    public TechnologyIdentified(String variableName, String technologyName)
+    {
+        super(variableName);
```
klinki: What is `variableName` ?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -0,0 +1,146 @@
...
+        this.technologyName = technologyName;
+    }
+
+    /**
+     * Create the operation with the given technology name.
+     */
+    public static TechnologyIdentifiedWithName named(String technologyName)
+    {
+        return new TechnologyIdentified(technologyName);
```
klinki: Why is this method needed, when constructor is `public`? Is it supposed to be `public`?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -0,0 +1,146 @@
...
+        return this;
+    }
+
+    /**
+     * Specifies a tag to associate with this technology.
+     */
+    public TechnologyIdentified withTag(String tag)
+    {
+        this.tags.add(tag);
+        return this;
```
klinki: Maybe rename this method to `addTag`? It is not clear `withTag` adds tag to collection.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -0,0 +1,146 @@
...
+        {
+            project = ((FileLocationModel)payload).getFile().getProjectModel();
+        }
+        else
+        {
+            return;
+        }
+
+        TechnologyUsageStatisticsService service = new TechnologyUsageStatisticsService(event.getGraphContext());
+        TechnologyUsageStatisticsModel model = service.get(project, this.technologyName);
```
klinki: Maybe rename `get` method to `getOrCreate` (as other methods below), since it creates new `TechnologyUsageStatisticsModel` if it doesn't exist.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyUsageStatisticsService.java
```diff
@@ -0,0 +1,384 @@
...
+        // technologyUsage.put(TechnologyUsageStatisticsModel.STATS_JAVA_CLASSES_TOTAL, this.countByType(JavaClassModel.class));
+        //
+        // // We are not able to tell which of the jars are original. We can substract known opensource libs.
+        // technologyUsage.put(TechnologyUsageStatisticsModel.STATS_JAVA_JARS_ORIGINAL, this.diff(
+        // this.countByType(JarArchiveModel.class),
+        // this.countByType(IdentifiedArchiveModel.class))
+        // );
+        // technologyUsage.put(TechnologyUsageStatisticsModel.STATS_JAVA_JARS_TOTAL, this.countByType(JarArchiveModel.class));
+
+        return this.groupByProjectModel(technologyUsage);
```
klinki: How would this method work? Is it obsolete now, or to be fixed later?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Adds the specified statistics and tag information regarding an technology that has been located in the analyzed application.
+ *
+ * If no count is specified, then a default of "1" is assumed.
+ *
+ * @author <a href="mailto:jesse.sightler@gmail.com">Jesse Sightler</a>
+ */
+public class TechnologyIdentified extends AbstractIterationOperation<WindupVertexFrame>
+            implements TechnologyIdentifiedWithCount, TechnologyIdentifiedWithName
```
klinki: How can this implement `TechnologyIdentifiedWithCount` and `TechnologyIdentifiedWithName` if both of these 2 interfaces have incompatible methods `withTag` and don't extend each other?

```java
public interface TechnologyIdentifiedWithName extends OperationBuilder
{
    TechnologyIdentifiedWithName withTag(String tag);
    TechnologyIdentifiedWithCount numberFound(int count);
}
```
```java
public interface TechnologyIdentifiedWithCount {
    TechnologyIdentifiedWithCount withTag(String tag);
}
```

(I got it, these methods return instance of `TechnologyIdentified` which implements both of these interfaces, but shouldn't there be some relation between them? Maybe one should extend other?)

Btw aren't names of these interfaces opposite? I would expect the `TechnologyIdentifiedWithCount` to have some method with count (`numberFound(int count);`) 

Also the `TechnologyIdentifiedWithName` should have something to do with name. It only has `withTag` so maybe it should be `TechnologyIdentifiedWithTag` ?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    /**
+     * Sets the number of items that have been found by this operation.
+     *
+     * This is optional, and if left unspecified, it will default to {@link TechnologyIdentified#DEFAULT_COUNT}.
+     */
+    public TechnologyIdentifiedWithCount numberFound(int count)
+    {
+        this.count = count;
```
klinki: Maybe rename to `setNumberFound` ?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyUsageStatisticsService.java
jsight: Because this is just to keep it from being null. There is no way to tell what the value should be from this point in the code and it doesn't really make sense to pass in an initial value, IMO.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyUsageStatisticsService.java
jsight: Yes, this is obsolete. I left the commented out code in place in case it is useful for when the rules are created. Ideally, the rules should be in xml, IMO, rather than something that requires code changes.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
jsight: This is in keeping with the pattern used by JavaClass, Hint, and other fluent APIs in our codebase.

I am fine with moving away from this pattern, but I kept it here. The naming conventions for the classes follows those naming conventions as well. We can discuss it at the standup if people would like to change that, though.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
jsight: It is just to have a fluent API. See above the comment regarding fluent APIs.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
jsight: It is just to have a fluent API. See above the comment regarding fluent APIs.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
jsight: It is just to have a fluent API. See above the comment regarding fluent APIs.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
jsight: +1 - I have made this change. I hate a method called "get" that actually creates, and yet I've managed to make that mistake myself. Oops.

Good catch! :)

PR #1169, opened 2017-09-08
QueryHandler: Specify some basic graph queries via XML

jsight:
This is an experiment that might make it easier to implement some of the technology report rules as xml. What do you think?

### DIFF
--- a/config-xml/tests/src/test/resources/query-handler.rhamt.xml
+++ b/config-xml/tests/src/test/resources/query-handler.rhamt.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0"?>
+<test xmlns="http://windup.jboss.org/schema/jboss-ruleset" id="query-handler-test-1">
+
+    <graph-query discriminator="FileResource" from="test-from" as="test-as">
+
+    </graph-query>
+</test>
+
--- a/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
+++ b/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
@@ -13,6 +13,7 @@ import org.jboss.windup.graph.model.WindupVertexFrame;
 import org.jboss.windup.graph.model.resource.FileModel;
 import org.jboss.windup.reporting.model.TagSetModel;
 import org.jboss.windup.reporting.service.TagSetService;
+import org.jboss.windup.rules.apps.javaee.model.JNDIResourceModel;
 import org.jboss.windup.rules.apps.javaee.model.stats.TechnologyUsageStatisticsModel;
 import org.jboss.windup.util.Logging;
 import org.ocpsoft.rewrite.context.EvaluationContext;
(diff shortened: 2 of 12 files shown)


mareknovotny:
@jsight  Could you explain more the background and problem this solves?

I can imagine that Technology stats ruleset is in XML, but more information helps to get the whole picture

jsight:
The related PR may be clarifying: https://github.com/windup/windup-rulesets/pull/262

config-xml/tests/src/test/resources/query-handler.rhamt.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0"?>
+<test xmlns="http://windup.jboss.org/schema/jboss-ruleset" id="query-handler-test-1">
+
+    <graph-query discriminator="FileResource" from="test-from" as="test-as">
```
mareknovotny: I miss the xsd definition for the graph-query.

jsight: Fixed.

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -110,41 +111,48 @@ public int getCount()
...
-            project = ((FileLocationModel)payload).getFile().getProjectModel();
+            projects.add(((FileLocationModel)payload).getFile().getProjectModel());
+        } else if (payload instanceof JNDIResourceModel)
+        {
+            JNDIResourceModel jndiResourceModel = (JNDIResourceModel)payload;
+            jndiResourceModel.getApplications().forEach(projects::add);
         }
         else
         {
             return;
```
klinki: @jsight Should we create `else if` branch for each model used for technology report? Because right now, writing rules for different models then those you have listed there won't have any effect. It will always end up in 
```java
else { return }
```
branch.

If so, is the only purpose of this code to get project model of the entity? There is method `getRootProjectModels` in `BelongsToProject` interface which is created exactly for this purpose

mareknovotny: it seems like that see @mrizzi PR i merged https://github.com/windup/windup/pull/1173

mrizzi: @klinki are you suggesting to use 
```java
((BelongsToProject) payload).getRootProjectModels().forEach(projects::add);
```
instead of the `if-else` chain?

klinki: Yes

klinki: I made PR: https://github.com/windup/windup/pull/1174 with BelongsToProject

jsight: Yeah, I think that makes sense too. I wasn't thinking about how many variations of project references there were when I added this.

Although, how does that work when the root project is "shared-libs"?

PR #1170, opened 2017-09-12
WINDUP-1677 inconsistent type value annotations

klinki:


graph/api/src/main/java/org/jboss/windup/graph/model/ArchiveModel.java
```diff
@@ -19,7 +19,7 @@
 @TypeValue(ArchiveModel.TYPE)
 public interface ArchiveModel extends FileModel
 {
-    String TYPE = "ArchiveModel:";
+    String TYPE = "ArchiveModel";
     String ARCHIVE_NAME = TYPE + "archiveName";
```
mareknovotny: you should have here colon you deleted in TYPE

mareknovotny:
the result is not still consistent, somewhere you removed colons from TYPE and they are missing in concatenated strings for complete properties.
The other things is that somewhere is used colons and somewhere dash.

graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationArchiveModel.java
```diff
@@ -2,7 +2,8 @@
 
 import com.tinkerpop.frames.modules.typedgraph.TypeValue;
 
-@TypeValue("ApplicationArchive")
+@TypeValue(ApplicationArchiveModel.TYPE)
 public interface ApplicationArchiveModel extends ApplicationModel, ArchiveModel
 {
+    String TYPE = "ApplicationArchiveModel";
```
OndraZizka: I would generally remove the Model suffix, it's not really needed and makes everything longer - esp. the JSON data, URLs etc.

graph/api/src/main/java/org/jboss/windup/graph/model/ArchiveModel.java
OndraZizka: The colon just delimits the type discriminator from edge label. Could be -, which makes less troubles with indexes IIRC.

reporting/api/src/main/java/org/jboss/windup/reporting/model/InlineHintModel.java
```diff
@@ -17,7 +17,7 @@
 @TypeValue(InlineHintModel.TYPE)
 public interface InlineHintModel extends EffortReportModel, FileLocationModel, TaggableModel
 {
-    String TYPE = "Hint";
+    String TYPE = "InlineHintModel";
```
OndraZizka: Same here. If there is 10000 hints, this adds 110 kb to the JSON data. Not sure if we care.

graph/api/src/main/java/org/jboss/windup/graph/model/ArchiveModel.java
mareknovotny: yep, dash or colon, if we prefer dash as it is more convenient for indexes, we should go with dashes everywhere, do you agree @OndraZizka ?

klinki: I don't understand why `ARCHIVE_NAME` has `TYPE` prefix, but other properties (`UNZIPPED_DIRECTORY, PARENT_ARCHIVE` don't. 

This seems like another inconsistency. Should there be `TYPE` prefix or not?

mareknovotny: That was the random thing to satisfy uniqueness, but here that goes as we have bunch of models, types, properties and checks for uniqueness needs to be done manually until somebody experience exceptions in runtime :(

jsight: +1, personally, I find the prefixes to be a bit distracting.

graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationArchiveModel.java
jsight: +1

PR #1171, opened 2017-09-12
WINDUP-1665 DiscoverWebXmlRuleProvider fix for geronimo-web-xml

mrizzi:
The issue:
1. The rule searches for "web-app" tag and if finds it in in the configuration file for geronimo `geronimo-web.xml` (it's not a regular `web.xml` file)
2. `geronimo-web.xml` has a `<naming:resource-ref>` definition, [this line from DiscoverWebXmlRuleProvider](https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/DiscoverWebXmlRuleProvider.java#L192) is executed 
3. but what is called `res-ref-name` in `web.xml`, in `geronimo-web.xml` is called `ref-name` so method [processElement from DiscoverWebXmlRuleProvider](https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/DiscoverWebXmlRuleProvider.java#L211) return an object without name field
4. then [GenerateJBossWebDescriptorRuleProvider](https://github.com/windup/windup/blob/master/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/jboss/GenerateJBossWebDescriptorRuleProvider.java) doesn't create a valid transformed file

The fix:

- avoid the rule to consider file named `geronimo-web.xml` because Geronimo always searches for a configuration file named this way (ref. [Apache Geronimo docs](http://geronimo.apache.org/GMOxDOC30/geronimo-webxml.html#geronimo-web.xml-Overview))
- add a test case `DiscoverWebXmlTest` to the rule using test data from the `web-xml` folder with a file to be found from the rule (`web.xml`) and one to be discarded (`geronimo-web.xml`)

### DIFF
--- a/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/DiscoverWebXmlRuleProvider.java
+++ b/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/DiscoverWebXmlRuleProvider.java
@@ -12,6 +12,7 @@ import org.jboss.windup.config.GraphRewrite;
 import org.jboss.windup.config.metadata.RuleMetadata;
 import org.jboss.windup.config.phase.InitialAnalysisPhase;
 import org.jboss.windup.config.query.Query;
+import org.jboss.windup.config.query.QueryPropertyComparisonType;
 import org.jboss.windup.config.ruleprovider.IteratingRuleProvider;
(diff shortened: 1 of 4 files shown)


windupgithubbot1:

PR #1173, opened 2017-09-13
WINDUPRULE-289 Adding EjbBeanBaseModel handling

mrizzi:
- added handling for `EjbBeanBaseModel` otherwise windup/windup-rulesets#263 wasn't able to add the technologies it finds
- after talking with @jsight, using `findAllByProperty(TechnologyUsageStatisticsModel.NAME, this.technologyName);` has better performance than `findAll();` that is still there commented to be used to debug in this phase of heavy creation of new rules


windupgithubbot1:

PR #1174, opened 2017-09-13
TechnologyIdentified - use BelongsToProject instead of if-else chain

klinki:
Instead of having huge if-else chain, I suggest to use `BelongsToProject` interface.

I think it will be more maintainable and readable.

### DIFF
--- a/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
+++ b/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
@@ -7,14 +7,11 @@ import java.util.logging.Logger;
 import org.jboss.windup.config.GraphRewrite;
 import org.jboss.windup.config.operation.Iteration;
 import org.jboss.windup.config.operation.iteration.AbstractIterationOperation;
-import org.jboss.windup.graph.model.FileLocationModel;
+import org.jboss.windup.graph.model.BelongsToProject;
 import org.jboss.windup.graph.model.ProjectModel;
 import org.jboss.windup.graph.model.WindupVertexFrame;
-import org.jboss.windup.graph.model.resource.FileModel;
 import org.jboss.windup.reporting.model.TagSetModel;
 import org.jboss.windup.reporting.service.TagSetService;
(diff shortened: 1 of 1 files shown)


mareknovotny:
yep, it seems like good idea, is the contract for BelongToProject documented as required?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -113,27 +110,20 @@ public int getCount()
...
+        if (payload instanceof BelongsToProject)
         {
-            EjbBeanBaseModel ejbBeanBaseModel = (EjbBeanBaseModel) payload;
-            ejbBeanBaseModel.getApplications().forEach(projects::add);
+            Iterable<ProjectModel> rootProjectModels = ((BelongsToProject) payload).getRootProjectModels();
+
+            for (ProjectModel projectModel : rootProjectModels)
+            {
+                projects.add(projectModel);
+            }
```
mrizzi: As discussed, you can replace it with 
```
((BelongsToProject) payload).getRootProjectModels().forEach(projects::add);
```

klinki: You are right. Are there any performance drawbacks of this approach? Because this one is easier to debug, if you need to see which projects are added.

mareknovotny: well, for such simple case lambda is fine and readable enough I think. if you need to debug it that is the different thing and could be a little harder

jsight: I'm still concerned that this is going to attach everything to the shared-libs project in the case of a shared jar.

jsight: I wrote a unit test and confirmed that it returns only the shared-libs project in the case of a library in shared libs.

Updated PR here: https://github.com/windup/windup/pull/1178

PR #1175, opened 2017-09-13
WINDUPRULE-286 Models to handle markup technologies

mrizzi:
- added *Models to handle html, css and js files
- added `MarkupFileMappingRuleProvider` to map them all in just one step


windupgithubbot1:

PR #1176, opened 2017-09-14
WINDUP-1653: fixed anchor in file link

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1653

### DIFF
--- a/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/freemarker/RenderLinkDirective.java
+++ b/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/freemarker/RenderLinkDirective.java
@@ -153,7 +153,7 @@ public class RenderLinkDirective implements WindupFreeMarkerTemplateDirective
         if (result == null)
             writer.write(linkText);
         else
-            renderLink(writer, cssClass, project, result.getReportFilename() + "#" + anchor, linkText);
+            renderLink(writer, cssClass, project, result.getReportFilename(), anchor, linkText);
     }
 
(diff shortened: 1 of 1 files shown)


rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/freemarker/RenderLinkDirective.java
```diff
@@ -232,7 +232,7 @@ private void renderLink(Writer writer, String cssClass, ProjectModel project, St
         if (cssClass != null)
             writer.append(" class='" + cssClass + "'");
         writer.append(" href='").append(href);
-        appendProject(writer, project);
+//            appendProject(writer, project);
```
mareknovotny: this doesn't seem to be used anywhere on reports and breaks the HardcodedIP report links to source report of the file occurence

jsight: (this is a followup on the standup discussion this morning, just to document the issue)

This is used in the following places:
https://github.com/windup/windup/blob/master/reporting/impl/src/main/resources/reports/templates/source.ftl#L216

https://github.com/windup/windup/blob/master/reporting/impl/src/main/resources/reports/templates/source.ftl#L24

It works around the fact that a single source file may be used in multiple projects. So depending upon which application was linked from, the source report may have to render a different navigation bar (because otherwise, the navigation bar links would not go to the right place).

mareknovotny:
retest

jsight:
This looks good to me and seems to work well.

PR #1177, opened 2017-09-14
WINDUPRULE-285 Rule Providers to identify MVC libraries

mrizzi:
The approach for these rules has been adding a specific tag for each one of the 4 libraries (Spring MVC, Struts, Wicket and GWT). They are working in an "on/off" way to say if a library is there or not (not counting the number of jars from the same library).
It has been mandatory to add another `else if` in `TechnologyIdentified` class to handle `TechnologyTagModel` interface that can not extend `HasApplications` (and hence `BelongsToProject`) because otherwise it introduces a circular dependency (from IDE `Adding dependency on module 'windup-rules-java-ee' will introduce circular dependency between modules 'windup-reporting' and 'windup-reporting-api'.`)

### DIFF
--- a/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
+++ b/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
@@ -12,6 +12,7 @@ import org.jboss.windup.graph.model.ProjectModel;
 import org.jboss.windup.graph.model.WindupVertexFrame;
 import org.jboss.windup.graph.model.resource.FileModel;
 import org.jboss.windup.reporting.model.TagSetModel;
+import org.jboss.windup.reporting.model.TechnologyTagModel;
 import org.jboss.windup.reporting.service.TagSetService;
 import org.jboss.windup.rules.apps.javaee.model.EjbBeanBaseModel;
 import org.jboss.windup.rules.apps.javaee.model.JNDIResourceModel;
@@ -131,6 +132,9 @@ public class TechnologyIdentified extends AbstractIterationOperation<WindupVerte
         {
             EjbBeanBaseModel ejbBeanBaseModel = (EjbBeanBaseModel) payload;
             ejbBeanBaseModel.getApplications().forEach(projects::add);
+        } else if (payload instanceof TechnologyTagModel)
+        {
+            ((TechnologyTagModel)payload).getFileModels().forEach(fileModel -> fileModel.getRootProjectModels().forEach(projects::add));
         }
         else
         {
--- a/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/DiscoverEmbeddedGwtLibraryRuleProvider.java
+++ b/rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/rules/DiscoverEmbeddedGwtLibraryRuleProvider.java
@@ -1,0 +1,57 @@
+package org.jboss.windup.rules.apps.javaee.rules;
+
+import org.jboss.windup.config.AbstractRuleProvider;
+import org.jboss.windup.config.GraphRewrite;
+import org.jboss.windup.config.loader.RuleLoaderContext;
+import org.jboss.windup.config.metadata.RuleMetadata;
+import org.jboss.windup.config.operation.iteration.AbstractIterationOperation;
+import org.jboss.windup.config.phase.InitialAnalysisPhase;
+import org.jboss.windup.config.query.Query;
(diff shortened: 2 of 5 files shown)


windupgithubbot1:

mareknovotny:
We could move **HasApplications** into rules-java-api module as I don't think the reason it is in **rules-java-ee** is clear to me.

jsight:
@mareknovotny - It seems like it could be moved to graph-api as well.

mareknovotny:
yep. more general module the better

jsight:
The doubt that I have is in regards to making these as Java rules. It seems that these could just use the "File" condition instead?

mareknovotny:
@jsight +1, the conditions are simple, so java rules impls are not needed.

mrizzi:
I'm sorry but i have been misled by looking at other `DiscoverEmbedded*LibraryRuleProvider` and i started creating these Java rules but XML ones are more suitable for these cases, i agree with you.

jsight:
@mrizzi - It doesn't help that I had forgotten about the simple "File" condition earlier.

mrizzi:
I've been doing some tests about this PR and i want to share some considerations/ideas:

- switching from Java to XML rules means 
  - not use anymore the `JarArchiveModel` in favor of `FileModel` that, as far as i understand, can have some performance drawbacks on large application analysis
  - not having the tags displayed in the `index` report page because Java rule adds `TechnologyTagModel` to `FileModel` while XML rule can add `<tag>` to `<classification>`
- using only XML rules leads to choose between two different approaches:
    - all in one XML rule to discover the framework _X_ and then add classification and information, e.g.
    ```xml
     <when>
         <file filename="{*}wicket{*}.jar"/>
     </when>
    <perform>
        <classification title="Embedded library Apache Wicket" category-id="optional" effort="0">
              <description>The application embeds an Apache Wicket library.</description>
             <tag>Apache Wicket (embedded)</tag>
        </classification>
        <technology-identified name="Wicket">
             <tag name="View"/>
             <tag name="Embedded"/>
             <tag name="MVC"/>
        </technology-identified>
    <perform>
    ```
    - two different rules: one for discovering the framework _X_ and the other one to add the technology information
    `discovering rule`
    ```xml
        <rule id="mvc-00001">
            <when>
                <file filename="{*}wicket{*}.jar"/>
            </when>
            <perform>
                <classification title="Embedded library Apache Wicket" category-id="optional" effort="0">
                    <description>The application embeds an Apache Wicket library.</description>
                    <tag>Apache Wicket (embedded)</tag>
                </classification>
            </perform>
        </rule>
    ```
    `technology rule`
    ```xml
            <when>
                <classification-exists classification="Embedded library Apache Wicket" />
            </when>
            <perform>
                <technology-identified name="Wicket">
                    <tag name="View"/>
                    <tag name="Embedded"/>
                    <tag name="MVC"/>
                </technology-identified>
            </perform>
    ```
    using the `classification` as a way to bridge information. This approach comes from the case when we have the `discovering` rule in Java and we have to create the `technology` rule in XML. 
Having two rules also implies that to test the `technology rule` you have to execute also the `discovering rule` so changing the [windup-jboss-ruleset.xsd](https://github.com/windup/windup/blob/master/config-xml/schema/windup-jboss-ruleset.xsd#L15) to have unbounded `<rulePath>` tags in `<ruletest>` and so being able to have a test file starting this way:
    ```xml
    <ruletest id="technology-usage-X-test" xmlns="http://windup.jboss.org/schema/jboss-ruleset">
        <testDataPath>data/X/</testDataPath>
        <rulePath>../X.windup.xml</rulePath> <!-- discovering rule -->
        <rulePath>../X-technology-usage.windup.xml</rulePath> <!-- technology rule -->
    ``` 

I think the first approach is more straightforward while the second is a cleaner approach but that's just my idea.
WDYT?

jsight:
@mrizzi @mareknovotny - For the first part of your issue, I think that we should prioritize fitting WINDUP-1627 into this sprint. It looks like that could be as simple as adding an enumeration to the File condition (Archive|File) to allow it to initially filter by type. Can we talk about that on Monday?

I have also filed WINDUP-1682. Maybe we can fit that into this sprint as well?

As to the last part, I tend to like the second option if there are no hints or anything to attach to the technology itself. If the hint and classification information will be extensive, then separating them into more appropriate places for each bit seems like the better approach.

mrizzi:
It sounds good to have both the issues in this sprint to me: they'll make a lot easier and better-performing writing rules for technologies.

For the second part, i think we can go with with second approach (2 different rules: `discovery` and `technology`) so that we'll have a set of "technology-only" rules in `technology-usage/` folder that we know are fed by discovery rules, no matter if XML or Java.

mrizzi:
PR windup/windup#1180 made obsolete this PR.

PR #1178, opened 2017-09-14
Updated TechnologyIdentified to a generic approach and added a unit t…

jsight:
…est.

Note, this targets the issue where finding out what App Modules an item belongs to. It's an improvement on BelongsToProject.

Previously, getApplications() was doing the same thing as getRootProjectModels(). Also, getRootProjectModels() was returning
only the shared-libs application in the case of a jar found in multiple applications.

This update adds a unit test to confirm this behavior and also makes the code attempt to associate the technology
statististics at the lowest possible module level. This help when we (eventually) enable filtering support.


jsight:
```
[jsightler@jsightlerredhatlaptop2 windup-web]$ 
Binary file addons/web-support/impl/target/classes/org/jboss/windup/web/addons/websupport/rest/graph/AbstractGraphResource.class matches
addons/web-support/impl/src/main/java/org/jboss/windup/web/addons/websupport/rest/graph/AbstractGraphResource.java:            return projectService.getRootProjectModels();
[jsightler@jsightlerredhatlaptop2 windup-web]$
```


It seems that this method isn't directly used in windup-web?

klinki:
Oh, sorry I was wrong. Actually only `belongsToProject` method is used.

graph/api/src/main/java/org/jboss/windup/graph/model/ProjectModel.java
```diff
@@ -225,12 +225,22 @@
...
+     * Note: This may be the synthetic shared-libs project in some cases.
+     *
      * @return ProjectModel representing the whole application
      */
     @JavaHandler
     ProjectModel getRootProjectModel();
 
     /**
+     * Returns all applications that this project is a part of. This could be multiple applications if this project
+     * is included multiple times.
```
klinki: So it means something like "get parent applications"? 
`getApplications()` method sort of implies these would be child applications.

graph/api/src/main/java/org/jboss/windup/graph/model/ProjectModel.java
```diff
@@ -256,6 +266,33 @@ public ProjectModel getRootProjectModel()
...
+
+            ProjectModel parent = project.getParentProject();
+            if (parent != null)
+                parent.getApplications().forEach(applications::add);
+
+            if (parent == null)
+                applications.add(project);
+
+            return applications;
+        }
```
klinki: Naming is really confusing :( 

Do I understand it correctly, that in case of this structure:
```
A
|--- B
|--- C
```
getApplications() called on A would return A,  called on B would return B and A?

rules-java-ee/addon/src/main/java/org/jboss/windup/rules/apps/javaee/TechnologyIdentified.java
```diff
@@ -113,27 +113,30 @@ public int getCount()
...
+            Iterable<ProjectModel> rootProjectModels = ((HasApplications) payload).getApplications();
+
+            for (ProjectModel projectModel : rootProjectModels)
+            {
+                projects.add(projectModel);
+            }
         }
         else
         {
+            LOG.warning("Unrecognized payload for TechnologyIdentified. Payload must be instance of BelongsToProject.");
```
mrizzi: I think `BelongsToProject` reference is now out-of-date with these changes.

graph/api/src/main/java/org/jboss/windup/graph/model/resource/FileModel.java
```diff
@@ -222,11 +222,21 @@ OnParseError fromName(String name){
...
 
     /**
+     * Gets the list of all applications that this file is a part of. This will include both the "shared-libs" project
+     * as well as any actual applications.
+     */
+    @JavaHandler
+    @Override
+    Iterable<ProjectModel> getApplications();
+
+    /**
```
klinki: Again, I find this quite confusing.

```
A
|--- B
|--- C
```
If I call `getApplication()` on A, what should I get? A? And for `getApplications()` result would be also A?
If I call `getApplication()` on B, what should I get? B? And for `getApplications()` result would be A, B?


(more complicated example with shared libs:)
```
A
|--- B
|    |--- C
|    |    |---D
|--- E
|    |--- C
|    |    |---F
```
Now we have 2 different instances of C. Lets call them C1 and C2.

If I call `getApplication()` on C1, result should be B, right? And for `getApplications()` result should be A, B, C1?
If I call `getApplication()` on C2, what should I get? E? And for `getApplications()` result would be A, E, C2?

And for D, what is result of `getApplication()`? `shared-libs` ? And `getApplications()` is A, B, C1, shared-libs?
And for F, what is result of `getApplication()`? `shared-libs` ? And `getApplications()` is A, E, C2, shared-libs?

graph/api/src/main/java/org/jboss/windup/graph/model/HasApplications.java
```diff
@@ -1,30 +1,27 @@
...
  */
 public interface HasApplications extends BelongsToProject
 {
+    /**
+     * Gets all root project models for current model (This will be mostly 1, but there are few exceptions which have multiple project models, so it
+     * returns Iterable to keep interface consistent)
+     *
+     * @return root project models
+     */
     Iterable<ProjectModel> getApplications();
```
klinki: If I understand it correctly and it really returns the parent project models, then the name of interface probably shouldn't be `HasApplications` but `IsPartOfApplications` or something like that.

`getApplications()` in combination with `HasApplications` interface would imply it returns child applications, not parent ones.

graph/api/src/main/java/org/jboss/windup/graph/model/ProjectModel.java
OndraZizka: +1. `getCointainingApplications()` ?

jsight: I'm just adding a few notes from the discussion from this morning for reference.

getApplications() is called that as there is no notion of "child" applications in Windup. The concept of an application is that the application is the root project and thus would never have a parent (by definition).

graph/api/src/main/java/org/jboss/windup/graph/model/ProjectModel.java
jsight: I'm just adding a few notes from the discussion from this morning for reference.

Both would return "A" as the input application in this case is "A".

graph/api/src/main/java/org/jboss/windup/graph/model/resource/FileModel.java
jsight: I'm just adding a few notes from the discussion from this morning for reference.

In the first case, it would always return A. In the second case, any call for "C" would return both "A" and the synthetic "shared-libs" application.

Calls for "D" would do exactly the same thing (return "A" and the synthetic shared-libs application).

graph/api/src/main/java/org/jboss/windup/graph/model/HasApplications.java
jsight: It isn't always accurate that it is a part of an application, though. For example, if we find a reference to a datasource in an application, that means that datasource was used by that application. It doesn't necessarily follow that the datasource is defined there.

HasApplications just implies that a link should be present, not that an ownership relationship necessarily exists.

jsight:
Merging for now... further name changes can come in a separate PR if needed.

PR #1180, opened 2017-09-20
WINDUP-1682 WINDUPRULE-285 WINDUP-1687 WINDUP-1688

mrizzi:
This PR fixes:

- [WINDUP-1682](https://issues.jboss.org/browse/WINDUP-1682) "_Create a TechnologyTag operation as well as handlers so that this can be supported in xml rules_"
- [WINDUP-1687](https://issues.jboss.org/browse/WINDUP-1687) "_windup-jboss-ruleset.xsd: wrong "lineitem-exists" definition_"
- [WINDUP-1688](https://issues.jboss.org/browse/WINDUP-1688) "_Multiple <rulePath> tags in XML tests_"

This PR is related to:
- #1177 because this PR makes #1177 obsolete and hence to be closed without merging it
- windup/windup-rulesets#267 because new XML rules can be created to discover libraries usage.

This PR contains:
- new `<technology-tag>` for `<perform>` operation in XML rules with
  - text within the tag used as technology name (e.g. `Apache Wicket (embedded)`, `Spring MVC (embedded)`)
  - attribute `level` to set the [`TechnologyTagLevel`](https://github.com/windup/windup/blob/master/reporting/api/src/main/java/org/jboss/windup/reporting/model/TechnologyTagLevel.java) (default is `INFORMATIONAL`)
- new `<technology-tag-exists>` to be used for testing the `<technology-tag>` tag with:
  - `technology-tag` attribute (required) to set with the name of the technology to search for (e.g. `Apache Wicket`)
  - `in` attribute to only consider entries that reference a file with the file name given in this attribute
- handlers and operations for the new tags
- tag `<rulePath>` from XML rule tests can be used multiple times in the same test in order to be able to execute more rules inside the same analysis (e.g. a rule for discovering a library and then the rule to associate a technology usage for technology report)

### DIFF
--- a/config-xml/schema/windup-jboss-ruleset.xsd
+++ b/config-xml/schema/windup-jboss-ruleset.xsd
@@ -12,7 +12,7 @@
             <xs:sequence>
                 <xs:element name="testDataPath" minOccurs="1" type="xs:string" />
                 <xs:element name="sourceMode" minOccurs="0" maxOccurs="1" type="xs:boolean" />
-                <xs:element name="rulePath" minOccurs="0" type="xs:string" />
+                <xs:element name="rulePath" minOccurs="0" maxOccurs="unbounded" type="xs:string" />
                 <xs:element name="source" minOccurs="0" maxOccurs="1" type="xs:string" />
                 <xs:element name="target" minOccurs="0" maxOccurs="1" type="xs:string" />
                 <xs:element ref="ruleset" minOccurs="0" maxOccurs="unbounded" />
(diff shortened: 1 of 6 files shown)


mareknovotny:
The only thing I forgot to ask is why TechnologyTag needs to have a level. I think the technology is always at informational level and I can't think of what type of technology could be at important level.

jsight:
@mareknovotny I think that we still have a few that are marked "IMPORTANT" from the java rules. It pretty much just sets it to red for cases where the expectation is that migration effort will be required (for example, deployment descriptors).

mrizzi:
+1
It only changes the color (grey/red) of the background of the tag box.
I also forgot to write that the `<technology-tag>` are showed in index page of the static report.

mareknovotny:
i know that levels signals the importance on reports, but technology tags here is  just for statistics or do we use them on application list/dashboard too if they are not bound to classification/hint?

mrizzi:
They work exactly like `TechnologyTagModel` in Java rules to they're attached to a `FileModel` without any need to be bound to a classification/hint.

jsight:
Looks good to me.

PR #1186, opened 2017-09-26
Improve AbstractPathConfigurationOption's error messages.

OndraZizka:


jsight:
I don't personally consider it an improvement to switch to the passive voice here.

mareknovotny:
:) me neither, I think using MUST is signaling the importance more

OndraZizka:
Don't you think the "ERROR: " in front of it is enough?
Also, I believe that in the RH corporate standards, yelling at users is discouraged. If this is printed, it is one of 4 lines printed. I don't think the user is going to miss the importance of the message which tells him why Windup didn't run.

OndraZizka:
@bergerhoffer Andrea, WDYT - which is better?

$ windup.sh --input non-existent.file
ERROR: input must exist!

-- or -

$ windup.sh --input non-existent.file
ERROR: Option input must be an existing file: non-existent.file

jsight:
@OndraZizka - We can take a poll on Monday. :) Feel free to bring it up then.

jsight:
retest

jsight:
(I suspect that this broke tests, btw)

PR #1188, opened 2017-09-26
Rename ApplicationModel to ApplicationProjectModel, add ApplicationIn…

OndraZizka:
…putPathModel, and mark the right ProjectModels and FileModels with those types.

### DIFF
--- a/graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationArchiveModel.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationArchiveModel.java
@@ -3,7 +3,7 @@ package org.jboss.windup.graph.model;
 import com.tinkerpop.frames.modules.typedgraph.TypeValue;
 
 @TypeValue(ApplicationArchiveModel.TYPE)
-public interface ApplicationArchiveModel extends ApplicationModel, ArchiveModel
+public interface ApplicationArchiveModel extends ArchiveModel
 {
     String TYPE = "ApplicationArchiveModel";
 }
--- a/graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationProjectModel.java
+++ b/graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationProjectModel.java
@@ -3,10 +3,10 @@ package org.jboss.windup.graph.model;
 import com.tinkerpop.frames.Property;
 import com.tinkerpop.frames.modules.typedgraph.TypeValue;
 
-@TypeValue(ApplicationModel.TYPE)
-public interface ApplicationModel extends WindupVertexFrame
+@TypeValue(ApplicationProjectModel.TYPE)
+public interface ApplicationProjectModel extends ProjectModel
 {
-    String TYPE = "ApplicationModel";
+    String TYPE = "ApplicationProjectModel";
     public static final String PROPERTY_APPLICATION_NAME = "applicationName";
 
     @Property(PROPERTY_APPLICATION_NAME)
--- a/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/scan/provider/DiscoverFilesAndTypesRuleProvider.java
+++ b/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/scan/provider/DiscoverFilesAndTypesRuleProvider.java
@@ -3,10 +3,13 @@ package org.jboss.windup.rules.apps.java.scan.provider;
 import org.jboss.windup.config.AbstractRuleProvider;
 import org.jboss.windup.config.GraphRewrite;
 import org.jboss.windup.config.loader.RuleLoaderContext;
+import org.jboss.windup.config.operation.GraphOperation;
 import org.jboss.windup.config.phase.DiscoveryPhase;
 import org.jboss.windup.config.query.Query;
 import org.jboss.windup.config.query.QueryGremlinCriterion;
 import org.jboss.windup.config.query.QueryPropertyComparisonType;
+import org.jboss.windup.graph.model.ApplicationInputPathModel;
+import org.jboss.windup.graph.model.ApplicationProjectModel;
 import org.jboss.windup.graph.model.WindupConfigurationModel;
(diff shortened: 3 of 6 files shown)


graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationArchiveModel.java
```diff
@@ -3,7 +3,7 @@
 import com.tinkerpop.frames.modules.typedgraph.TypeValue;
 
 @TypeValue(ApplicationArchiveModel.TYPE)
-public interface ApplicationArchiveModel extends ApplicationModel, ArchiveModel
+public interface ApplicationArchiveModel extends ArchiveModel
```
jsight: Why would this no longer extend ApplicationModel?

jsight: Actually, I see why, but this all seems a bit strange now. It seems to me that ApplicationModel is a marker interface to indicate that something is an application.

Previously, (filemodel instanceof ApplicationModel) == true for cases where the file was an application.

After this change, that relationship is broken as you have redefined the meaning of "ApplicationModel".

graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationProjectModel.java
```diff
@@ -3,10 +3,10 @@
 import com.tinkerpop.frames.Property;
 import com.tinkerpop.frames.modules.typedgraph.TypeValue;
 
-@TypeValue(ApplicationModel.TYPE)
-public interface ApplicationModel extends WindupVertexFrame
+@TypeValue(ApplicationProjectModel.TYPE)
+public interface ApplicationProjectModel extends ProjectModel
```
jsight: This isn't just changing a name, this is changing the meaning, and it seems to me that the new meaning is less useful.

Why not just add the ApplicationModel marker interface to a ProjectModel vertex if you want to be able to differentiate them by type?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/scan/provider/DiscoverFilesAndTypesRuleProvider.java
```diff
@@ -32,9 +36,22 @@ public Configuration getConfiguration(RuleLoaderContext ruleLoaderContext)
     {
         return ConfigurationBuilder.begin()
 
+        // Mark all input paths as ApplicationProjectModel.
```
jsight: This comment is actually incorrect.

graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationArchiveModel.java
OndraZizka: Yes, it's an marker interface. But the name is ambiguous, so I am creating ApplicationProjectModel, which marks a ProjectModel that is an app, and a ApplicationInputPathModel, which marks what is now FileModel instanceof ApplicationModel.

OndraZizka: Note that a directory may also be an application, while not being an archive.

graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationProjectModel.java
OndraZizka: See above. I want to be able to ask for an app input paths or app projects, separatedly. Just marking them would not allow that. Plus, using it as general marker interface closes the possibility to have Project or File specific methods, relations and properties.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/scan/provider/DiscoverFilesAndTypesRuleProvider.java
OndraZizka: Right, this was changed by IDEA refactoring... fixed

graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationProjectModel.java
jsight: With the marker interface, you could have done "new GraphService<ApplicationModel>(event, ApplicationModel.class).findAll()". Now, that is impossible, so I don't see how getting rid of the ApplicationModel marker has helped you.

jsight:
Instead of this, I would suggest the following hierarchy:

  - ApplicationFileModel
     - ApplicationArchiveModel
- ProjectModel
  - ApplicationProjectModel

In this way, ApplicationModel can still be a marker interface for all Application File objects and ProjectModel has a separate hierarchy.

The existing ApplicationModel would be renamed to ApplicationFileModel as that is really what it is. From what I can tell, the "ambiguity" that you are imaginging is the absence of an ApplicationProjectModel concept.

graph/api/src/main/java/org/jboss/windup/graph/model/ApplicationProjectModel.java
OndraZizka: What is that good for? It is never used anywhere.

OndraZizka:
Could be, perhaps. Isn't `ApplicationInputPath` a better matching name? It can be a directory, too. We don't even know if it is an existing file. We only know it's a path, initially.

jsight:
@OndraZizka - Maybe ApplicationInputPathModel is ok, but is that always an accurate name? If the user gives us a directory as the input path and does not specify "exploded" then we will crawl the input directory and create applications from all files found there. Then we will have many application paths, but still only one InputPath. Wouldn't input path be misleading in that case? I agree that differentiating ApplicationArchive from ApplicationInputPath is probably unnecessary.

I was horribly unclear in my last sentence above. ApplicationFileModel (or ApplicationInputPathModel) would take the place of the existing ApplicationModel.

OndraZizka:
Hmm good point. I'll think about it :)

PR #1189, opened 2017-09-26
XmlFileXpathValidator - create the snippet properly - just from the n…

OndraZizka:
…ode, not it's child nodes. Plus some small refactoring.


jsight:
What problem does this solve?

OndraZizka:
1) It doesn't show irrelevant content in places where the snippet is displayed. 
2) Excessive log output, making it hard to find the relevant information in it.

jsight:
I will need to see some examples of where that happens in order to understand that.

OndraZizka:
Check the documentation of the original length code. It gives a description of the element object rather than the XML code as assumed in the code. It's a buggy code.
Then the second part is that as the snippet, the whole XML tree is taken. My change only takes the relevant element's serialization length. Not perfect, not exact, but better than the basically random length or the wole XML subtree currently used.

windupgithubbot1:

windupgithubbot1:

windupgithubbot1:

PR #1194, opened 2017-09-26
Log msg changes

OndraZizka:
This makes the log a bit more readable.

### DIFF
--- a/reporting/api/src/main/java/org/jboss/windup/reporting/config/Hint.java
+++ b/reporting/api/src/main/java/org/jboss/windup/reporting/config/Hint.java
@@ -27,6 +27,7 @@ import org.jboss.windup.reporting.category.IssueCategoryModel;
 import org.jboss.windup.reporting.category.IssueCategoryRegistry;
 import org.jboss.windup.util.ExecutionStatistics;
 import org.jboss.windup.util.Logging;
+import org.jboss.windup.util.Util;
 import org.ocpsoft.rewrite.config.OperationBuilder;
 import org.ocpsoft.rewrite.config.Rule;
 import org.ocpsoft.rewrite.context.EvaluationContext;
@@ -189,7 +190,7 @@ public class Hint extends ParameterizedIterationOperation<FileLocationModel> imp
                 ((SourceFileModel) locationModel.getFile()).setGenerateSourceReport(true);
 
             LOG.info("Hint added to " + locationModel.getFile().getPrettyPathWithinProject()
-                    + ":\n    " + this.toString(hintModel.getTitle(), null));
+                    + ":"+ Util.NL +"    " + this.toString(hintModel.getTitle(), null));
         }
         finally
         {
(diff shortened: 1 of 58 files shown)


reporting/api/src/main/java/org/jboss/windup/reporting/config/Hint.java
```diff
@@ -284,13 +285,14 @@ private String toString(String title, String text)
...
         if (effort != 0)
             result.append(".withEffort(").append(effort).append(")");
-        if (links != null && !links.isEmpty())
-            result.append(".with(").append(links).append(")");
         if (tags != null && !tags.isEmpty())
             result.append(".withTags(").append(tags).append(")");
+        if (text != null)
+            result.append("\n\t.withText(\"").append(StringUtils.abbreviate(text.trim(), MAX_LOG_TEXT_LENGTH)).append("\")");
+        if (links != null && !links.isEmpty())
+            result.append("\n\t.with(").append(links).append(")");
```
mareknovotny: It is better to use `System.lineSeparator()` instead of `\n` to be the same on all OS

reporting/api/src/main/java/org/jboss/windup/reporting/config/Hint.java
```diff
@@ -187,8 +188,8 @@ public void performParameterized(final GraphRewrite event, final EvaluationConte
             if (locationModel.getFile() instanceof SourceFileModel)
                 ((SourceFileModel) locationModel.getFile()).setGenerateSourceReport(true);
 
-            LOG.info("Hint added to " + locationModel.getFile().getPrettyPathWithinProject() + " [" + this.toString(hintModel.getTitle(), hintText)
-                        + "] with tags: " + StringUtils.join(this.getTags(), " "));
+            LOG.info("Hint added to " + locationModel.getFile().getPrettyPathWithinProject()
+                    + ":\n    " + this.toString(hintModel.getTitle(), null));
```
mareknovotny: It is better to use `System.lineSeparator()` instead of `\n` to be the same on all OS.

reporting/api/src/main/java/org/jboss/windup/reporting/config/Hint.java
jsight: +1 - System.lineSeparator() is better.

OndraZizka: It doesn't really matter anymore, all Windows apps can handle \n. But as you like

reporting/api/src/main/java/org/jboss/windup/reporting/config/Hint.java
OndraZizka: Btw how does it matter if our logs are full of newlines coming from our rules?

OndraZizka: Bump. We have \n all over the project, so unless we are determined to fix all of them, I think there is no need to hold this PR because of that.

jsight: @OndraZizka - Fixing the PR is trivial. I agree that we should fix some of the other places where we do this, but that isn't a reason to add more places that need fixing. Just take the 5 minutes to fix and repush. :)

OndraZizka: Ok, but the code is then cluttered with unnecessary calls. The Windows world is now accepting `\n` just ok.

OndraZizka:
Rebased and now using the NL constant. Marek's PR not only makes the code harder to read, but also changed the whitespace of certain things. This is addressing the first issue.

PR #1196, opened 2017-09-28
FreeMarker - add few convenience variables and a little refactoring.

OndraZizka:


jsight:
How is this different from https://github.com/windup/windup/pull/1192 ?

OndraZizka:
Almost the same. I forgot about the other one. Now it's closed.

PR #1197, opened 2017-09-28
Add @JavaHandler retrieveSize() to FileModel.

OndraZizka:


jsight:
This really shouldn't be a separate PR, unless someone else is depending on it? Also, it should probably just be "getSize()" as an accessor method.

PR #1193, opened 2017-09-26
Rename <tag root="..."> attrib to <tag prime="...">

OndraZizka:
As discussed at the meeting.

PR #1200, opened 2017-09-28
Add TagModel and TagGraphService

OndraZizka:
This will be used in the reports.  Depends on #1193 .
(I am splitting the coming tech report PR into things that can be isolated.)


jsight:
Just include this in the tech report PR. If you really want it to be a standalone PR, it will need tests, IMO, as it is difficult to review code like this in complete isolation.

PR #1202, opened 2017-10-02
Do not fork bootstrap tests. These share forge addons and appear to be causing occasional race conditions

jsight:


mareknovotny:
@jsight Are you experiencing the timeouts locally too? I set jenkins jobs with -DforkCount=1 and it seems work -> no need to change the pom.xml ;) But the persistent change is good to have too.

jsight:
No, I didn't have the failures locally, but I don't run all tests locally that often either. I think that it would be best if the tests could still fork on Wonka for the other modules than the test module.

PR #1208, opened 2017-10-12
WINDUP-1730: fixing validation of input/output options

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1730

I created new test for faililing soon `NoInputOrOutputPathTest`

I am not able to shift Target prompt if none is given as it would fail on validation.

### DIFF
--- a/bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
@@ -458,19 +458,19 @@ public class Bootstrap
         return true;
     }
 
-    private File getWindupAddonsDir()
-    {
-        return new File(getWindupHome(), "addons");
-    }
+//    private File getWindupAddonsDir()
+//    {
+//        return new File(getWindupHome(), "addons");
+//    }
 
-    private File getWindupHome()
-    {
-        String windupHome = System.getProperty(WINDUP_HOME);
-        if (windupHome == null)
-        {
-            Path path = new File("").toPath();
-            return path.toFile();
-        }
-        return Paths.get(windupHome).toFile();
-    }
+//    private File getWindupHome()
+//    {
+//        String windupHome = System.getProperty(WINDUP_HOME);
+//        if (windupHome == null)
+//        {
+//            Path path = new File("").toPath();
+//            return path.toFile();
+//        }
+//        return Paths.get(windupHome).toFile();
+//    }
 }
--- a/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/RunWindupCommand.java
+++ b/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/RunWindupCommand.java
@@ -335,7 +335,8 @@ public class RunWindupCommand implements Command, FurnaceDependent
 
     private void setDefaultOutputPath(Map<String, Object> optionValues)
     {
-        if (!optionValues.containsKey(OutputPathOption.NAME))
+        Object obj = optionValues.getOrDefault(OutputPathOption.NAME, null);
+        if (obj == null || (obj instanceof File && StringUtils.isBlank( ((File) obj).getPath())) )
         {
             Iterable<Path> paths = (Iterable<Path>) optionValues.get(InputPathOption.NAME);
             if (paths != null && paths.iterator().hasNext())
--- a/config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
+++ b/config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
@@ -73,7 +73,7 @@ public abstract class AbstractPathConfigurationOption extends AbstractConfigurat
         if ( (fileObject == null && isRequired()) ||
              (fileObject instanceof Collection && isRequired() && ((Collection) fileObject).isEmpty()) )
         {
-            return new ValidationResult(ValidationResult.Level.ERROR, getLabel() + " must be specified.");
+            return new ValidationResult(ValidationResult.Level.ERROR, getName() + " must be specified.");
         }
         else if (fileObject == null)
         {
--- a/config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
+++ b/config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
@@ -3,6 +3,7 @@ package org.jboss.windup.config;
 import java.io.File;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Collection;
 
 import org.jboss.windup.util.exception.WindupException;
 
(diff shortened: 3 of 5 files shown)


bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
```diff
@@ -458,19 +458,19 @@ private boolean executePhase(CommandPhase phase, CopyOnWriteArrayList<Command> c
...
+//    private File getWindupHome()
+//    {
+//        String windupHome = System.getProperty(WINDUP_HOME);
+//        if (windupHome == null)
+//        {
+//            Path path = new File("").toPath();
+//            return path.toFile();
+//        }
+//        return Paths.get(windupHome).toFile();
+//    }
```
mareknovotny: Not related, but nowhere used so commenting them

bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/RunWindupCommand.java
```diff
@@ -335,7 +335,8 @@ private boolean validateOptionValues(Map<String, ConfigurationOption> options, M
 
     private void setDefaultOutputPath(Map<String, Object> optionValues)
     {
-        if (!optionValues.containsKey(OutputPathOption.NAME))
+        Object obj = optionValues.getOrDefault(OutputPathOption.NAME, null);
+        if (obj == null || (obj instanceof File && StringUtils.isBlank( ((File) obj).getPath())) )
         {
```
mareknovotny: If you launch:
`rhamt-cli --input something --output     --target eap`
this should avoid it and let set the default output directory

config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
```diff
@@ -65,12 +66,14 @@ else if (!Files.exists(path))
...
+    @SuppressWarnings("rawtypes")
     @Override
     public ValidationResult validate(Object fileObject)
     {
-        if (fileObject == null && isRequired())
+        if ( (fileObject == null && isRequired()) ||
+             (fileObject instanceof Collection && ((Collection) fileObject).isEmpty()) )
         {
-            return new ValidationResult(ValidationResult.Level.ERROR, getName() + " is required!");
+            return new ValidationResult(ValidationResult.Level.ERROR, getLabel() + " must be specified.");
```
mareknovotny: I think better to use label(s) in ValidationResult messages instead of name which is short.

exec/api/src/main/java/org/jboss/windup/exec/configuration/options/OutputPathOption.java
```diff
@@ -77,17 +77,19 @@ public static ValidationResult validateInputAndOutputPath(Path inputPath, Path o
...
 
         if (inputPaths == null || inputPaths.isEmpty())
         {
             return new ValidationResult(ValidationResult.Level.ERROR, "Input path must be specified.");
         }
+        
+        if (outputPath == null)
+        {
+            return new ValidationResult(ValidationResult.Level.ERROR, "Output path must be specified.");
+        }
```
mareknovotny: Switching the order of check, `inputPath` is more important than `outputPath`

mareknovotny:
retest

bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/RunWindupCommand.java
OndraZizka: ok

bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
OndraZizka: ok

config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
```diff
@@ -65,12 +66,14 @@ else if (!Files.exists(path))
...
     @Override
     public ValidationResult validate(Object fileObject)
     {
-        if (fileObject == null && isRequired())
+        if ( (fileObject == null && isRequired()) ||
+             (fileObject instanceof Collection && isRequired() && ((Collection) fileObject).isEmpty()) )
         {
-            return new ValidationResult(ValidationResult.Level.ERROR, getName() + " is required!");
+            return new ValidationResult(ValidationResult.Level.ERROR, getLabel() + " must be specified.");
         }
```
OndraZizka: I'm not sure about this. Isn't name what's used for the actual user input on command line?

exec/api/src/main/java/org/jboss/windup/exec/configuration/options/OutputPathOption.java
OndraZizka: ok

OndraZizka: Although in this case, it doesn't really matter because this is the API part, right? I.e. dev time error.
If it's not, then what's the handling of default at the top of this PR?

config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
mareknovotny: Well, there are mixed/not consistent validation messages. Somewhere it is used just strings not related to name/label.

exec/api/src/main/java/org/jboss/windup/exec/configuration/options/OutputPathOption.java
mareknovotny: Not really sure, what you are trying to say by that "API part".

config/api/src/main/java/org/jboss/windup/config/AbstractPathConfigurationOption.java
OndraZizka: Then probably we should make it consistent? Changing this will fix the inconsistency?

jsight: +1 - It should always display the parameter name, IMO.

mareknovotny: ok, then the mixed validation message in https://github.com/windup/windup/blob/master/exec/api/src/main/java/org/jboss/windup/exec/configuration/options/OutputPathOption.java#L132 and above about `--output` are little different too

jsight: I think some of those should display the parameter as well. Some of them are about logical errors, though, and the parameter is a little less important (since the user has clearly specified both already).

mareknovotny: if we are fine with such inconsistencies I will change it back.

PR #1210, opened 2017-10-19
Added an untar utility and a service provider interface for registering services (such as a JMS based executor service).

jsight:


jsight:
retest

server-provider-spi/pom.xml
```diff
@@ -0,0 +1,20 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.jboss.windup</groupId>
+        <artifactId>windup-parent</artifactId>
+        <version>4.1.0-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.jboss.windup</groupId>
```
mareknovotny: this is not needed, it is the same as parent groupId

utils/pom.xml
```diff
@@ -37,6 +37,13 @@
             <version>18.0</version>
         </dependency>
 
+        <!-- JTar - Java Tar file support -->
+        <dependency>
+            <groupId>org.kamranzafar</groupId>
+            <artifactId>jtar</artifactId>
+            <version>2.3</version>
+        </dependency>
```
mareknovotny: Is this library a good for the purpose? never heard about it ;)

mareknovotny:
retest

utils/pom.xml
jsight: Good catch, fixed!

server-provider-spi/pom.xml
jsight: Good catch, fixed!

utils/pom.xml
jsight: Yes, it is handy if you want to bundle a bunch of files up into a single archive, but you don't want to spend the CPU on compressing them. I could potentially have used the zip files for a similar purpose, but tar seems like a more suitable format for a simple concatenation of files.

It also leaves open the possibility of streaming them in the future, which isn't possible with zip.

mareknovotny:
it seems tests worked fine before update and now our CI triggered multiple times that queue is veeery long. I won't wait for this PR as it clearly works fine.

PR #1211, opened 2017-10-23
Windup 1743

mareknovotny:
https://issues.jboss.org/browse/WINDUP-1743

This is primary targeting 4.0.x branch. After merging into that this needs to be merged into master too.

### DIFF
--- a/rules-java/api/src/main/resources/reports/templates/ignored_files.ftl
+++ b/rules-java/api/src/main/resources/reports/templates/ignored_files.ftl
@@ -15,6 +15,9 @@
 
 <#macro ignoredFileRenderer reportModel>
     <div class="panel panel-primary">
+        <div class="panel-heading">
+            <h3 class="panel-title">Ignored files</h3>
+        </div>
         <table class="table table-striped table-bordered">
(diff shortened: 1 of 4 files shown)


mareknovotny:
tests are failing, but the target is for 4.0.x  branch so that is the reason

rules-java/api/src/main/resources/reports/templates/ignored_files.ftl
```diff
@@ -15,6 +15,9 @@
 
 <#macro ignoredFileRenderer reportModel>
     <div class="panel panel-primary">
+        <div class="panel-heading">
+            <h3 class="panel-title">Ignored files</h3>
```
jsight: Shouldn't "files" be upper-case ("Files")? I think that it may be wrong in a few places in this report.

mareknovotny: right, I didn't catch this.

mrizzi: Good catch, i missed this too :(

PR #1212, opened 2017-10-23
WINDUP-1746: DiscoverEmbeddedJSFLibraryRuleProvider should also find myfaces jars

jsight:


mareknovotny:
not sure if that should be here as  alternative to jsf discovered too, but myfaces has more than myfaces-api.jar or myfaces-impl.jar from version 2.x. https://myfaces.apache.org/core20/index.html
They have additionally to api, impl jars also myfaces-bundle.jar - see 

> The Bundle submodule provides both API and Impl classes into a single jar file with a OSGi MANIFEST.MF

.

jsight:
Good catch... added the bundle one as well. I think this can be merged once the tests finish.

It still bothers me that the guidance given by this rule is really incorrect, though. :( Filed that separately as: https://issues.jboss.org/browse/WINDUPRULE-318

PR #1215, opened 2017-10-30
WINDUP-1724 Reduce list {source,target,tag} commands execution time

mrizzi:
On average, an execution of RHAMT CLI with one of the following flags `--listTags`, `--listSourceTechnologies` or`--listTargetTechnologies` tooks about 5 seconds to execute.
In this 5 seconds, about 1.5-2 seconds are spent in a for loop in [XMLRuleProviderLoader](https://github.com/windup/windup/blob/master/config-xml/addon/src/main/java/org/jboss/windup/config/parser/XMLRuleProviderLoader.java#L120) due to loading of more than 100 rules files, each taking between 15-20 milliseconds.
The remaining 3 seconds are spent for starting and then shutting down Furnace and scanning the classpath for all Java addons.
With the changes in this PR, the sources/targets/tags are not retrieved  from rules anymore but they're read from the `help.xml` file with the result that running every command now take only 3 seconds (sometimes just 2) that is the minimum time for booting Furnace and then turn it off.

### DIFF
--- a/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/AbstractListCommand.java
+++ b/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/AbstractListCommand.java
@@ -1,11 +1,9 @@
 package org.jboss.windup.bootstrap.commands;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 
 import org.jboss.forge.furnace.Furnace;
+import org.jboss.windup.bootstrap.help.Help;
 
 /**
  * @author <a href="mailto:lincolnbaxter@gmail.com">Lincoln Baxter, III</a>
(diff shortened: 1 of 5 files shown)


jsight:
Hmm, why does it still have to boot Furnace?

mrizzi:
Commit 8d62b1c:

- changes `List{SourceTechnologies,TargetTechnologies,Tags}Command` execution phase from [`CommandPhase.PRE_EXECUTION`](https://github.com/windup/windup/blob/master/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/CommandPhase.java#L30) to [`CommandPhase.PRE_CONFIGURATION`](https://github.com/windup/windup/blob/master/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/CommandPhase.java#L15) in order to avoid executing Furnace (@jsight thanks)
- adds `AbstractListCommandWithoutFurnace` for commands that display list but do not need Furnace for their executions while the _old_ `AbstractListCommand` denotes commands that need Furnace to be executed (i.e. `DiscoverPackagesCommand` & `ListAddonsCommand`)

mrizzi:
retest

mrizzi:
retest

jsight:
Oh, nice, it finally passed. :)

Merging...

PR #1205, opened 2017-10-10
WINDUP-1684 tech report boxes

OndraZizka:
Working preview. Needs cleanup, tests, icons, squashing, could have some bugs.

PR #1222, opened 2017-11-08
Ondra zizka windup 1684 tech report boxes

jsight:
Followup on: https://github.com/windup/windup/pull/1205


jsight:
@mrizzi - I have made some tweaks to this that I believe make it a bit clearer. Can you take a look at it and do some testing?

reporting/api/src/main/java/org/jboss/windup/reporting/service/TagGraphService.java
```diff
@@ -0,0 +1,183 @@
...
+     * If the Tag was already processed exists, returns the corresponding TagModel.
+     * Doesn't check if the TagModel for given tag name already exists, assuming this method is only called once.
+     */
+    private TagModel feedTagStructureToGraph(Tag tag, Set<Tag> visited, int level)
+    {
+        if (visited.contains(tag))
+            return this.getUniqueByProperty(TagModel.PROP_NAME, tag.getName(), true);
+        visited.add(tag);
+
+        LOG.info(String.format("Creating TagModel for Tag: %s%s(%d)   '%s'  traits: %s", StringUtils.repeat(' ', level*2),
```
mrizzi: Can it be at `debug` level?

graph/api/src/main/java/org/jboss/windup/graph/service/GraphService.java
```diff
@@ -224,20 +247,37 @@ public T getUnique() throws NonUniqueResultException
...
-        if (iterator.hasNext())
-        {
-            throw new NonUniqueResultException("Expected unique value, but returned non-unique.");
+            WindupVertexFrame item = iterator.next();
+
+            // There can be other types using the same property name.
+            if (!type.isInstance(item))
+                continue;
+
+            if ("jsf".equals(value))            /// DEBUG, remove
```
mrizzi: I got this entry in the log file:
`[FOO] (main) Model with name 'jsf', type: interface org.jboss.windup.reporting.model.TagModel item: interface org.jboss.windup.graph.model.WindupVertexFrame, interface com.tinkerpop.frames.VertexFrame, interface org.jboss.windup.reporting.model.TagModel`
Is it useful to debug? Or can it be removed?

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -0,0 +1,348 @@
...
+                    .map(Long.class::cast)
+                    .collect(Collectors.toList());
+
+            LOG.info(String.format("--- Adding to projects %s: tech '%s', count: %sx, tags: %s", StringUtils.join(appsToCountTowards, " "),
+                    stat.getName(), stat.getOccurrenceCount(), stat.getTags()));
+
+            // A shortcut.
+            if (onlyID != null && !appsToCountTowards.contains(onlyID))
+            {
+                LOG.info("\t\tThis stat is not for this project, skipping.");
```
mrizzi: Can it be at `debug` level?

graph/api/src/main/java/org/jboss/windup/graph/service/GraphService.java
jsight: lol, how did I not notice that?

jsight:
@mrizzi - I think that I've fixed the story point issue.

I am confused by your last point, though. The sizes of the circles are determined vertically and not horizontally, right? So the size of the markup one would be correct, if the other ones have less "markup". The size of the IoC bubble would be completely unrelated to the "markup" bubble. I may have just misunderstood you here, though.

mrizzi:
@jsight you've not misunderstood me, i was comparing the bubbles horizontally because, at first saw, i tried to compare them vertically and the 2 `IoC` bubbles are the same but the `myproject-1.0-SNAPSHOT.war` has 2 `IoC` (both from `Spring` technology) and the `brownie.ear` has 18 `IoC` (17 from `Spring` and 1 from `AspectJ`) so i was confused too.
It seems to me the vertical comparisons are the right one but then the `IoC` bubbles are wrong.

mrizzi:
@jsight thanks: the mandatory SP issue is fixed 👍 
I tried also to compare the bubbles and boxes report across different browsers (Chrome and FF) and the boxes report is consistent between the browsers while the bubbles one has some issues regarding the bubbles size because some bubbles, that seem the same in FF, have different sizes in Chrome. See, for example, the bubbles in the `View` section, in the screenshots below:

_Firefox_
screenshot from 2017-11-10 10-24-30

_Chrome_
screenshot from 2017-11-10 10-24-42

jsight:
@mrizzi - I think that I've fixed the count issue. It wasn't handling models attached to projects correctly and was expecting them only to attach to applications.

I'm not really sure about the sizes, though. I'm wondering if there is a better approach available somehow vs the CSS text thing?

mrizzi:
@jsight the count is now fixed, thanks 👍 
On Chrome the bubbles report looks really better now (on FF is better too but the smallest bubble are rendered in a different way than in Chrome).

I think we can find a better way to create bubbles with the right size rather than the current CSS content but maybe it's out of scope for this PR since now the size algorithm is working and we can wait for the report to be tested before deciding in which way to enhance the current implementation
wdyt?

jsight:
@mrizzi - I think that is fine. We can file a couple of JIRAs for known issues remaining as well.

mrizzi:
For bubbles' alignment has been created [WINDUP-1771](https://issues.jboss.org/browse/WINDUP-1771)

PR #1223, opened 2017-11-08
WINDUP-1767 CLI to describe what parameters are required

mrizzi:
[`org.jboss.windup.config.ConfigurationOption#isRequired`](https://github.com/windup/windup/blob/master/config/api/src/main/java/org/jboss/windup/config/ConfigurationOption.java#L43) is now handled in the [`OptionDescription`](https://github.com/windup/windup/blob/master/bootstrap/src/main/java/org/jboss/windup/bootstrap/help/OptionDescription.java) so that it can be displayed in the output of the help command for every option.

### DIFF
--- a/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/DisplayHelpCommand.java
+++ b/bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/DisplayHelpCommand.java
@@ -22,7 +22,9 @@ public class DisplayHelpCommand implements Command
         for (OptionDescription option : Help.load().getOptions())
         {
             sb.append("--").append(option.getName()).append(System.lineSeparator());
-            sb.append("\t").append(option.getDescription()).append(System.lineSeparator());
+            sb.append("\t");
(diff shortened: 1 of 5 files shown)


mrizzi:
retest

jsight:
Looks good to me.

jsight:
I wonder if some reordering of the options could be helpful as well? --target is required now, but it is very far down the list, for example.

Similarly, --batchMode and --overwrite seem like commonly used features.

What do you think?

mrizzi:
I think you're right and some reordering is needed in the option list for the output of the help command so i made some changes and pushed them.

0579fa1 commit:
- changed the sorting algoritm for `ConfigurationOption` in `WindupConfiguration` so that `required` options are before the optional one in the help output (if both options are required or optional, then the priority is still the way to sort them)
- added priority to `OverwriteOption` to place it just behind the `--output` option
- `batchMode` is now the first one of the Forge options (i didn't move it before in the list to not mix RHAMT and Forge options)

jsight:
Also merged to 4.0.x

PR #1227, opened 2017-11-09
WINDUP-1759 - Update tattletale version

klinki:
TODO: 
- Need to release `1.3` GA version.


jsight:
I don't think that we should do this until a non-SNAPSHOT version is available. You can release that at any time, though.

klinki:
I just published non-SNAPSHOT version. Let's just wait until it is available and wonka build passes.

jsight:
Great! :)

PR #1233, opened 2017-11-14
WINDUP-1770 static reports performance

klinki:
Tested in Firefox and Chrome, seems to work smoothly

Note: This needs to be merged also to 4.0.x branch


jsight:
This looks great to me. @mrizzi - Feel free to rereview. I think it is ready to merge, though?

PR #1235, opened 2017-11-14
WINDUP-1774 CLI --output option handling

mrizzi:
The test didn't cover the right case because `""` was one of the element in the [`args`](https://github.com/windup/windup/blob/master/bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java#L72) array while if the command `rhamt-cli --input something --output --target eap` is executed there's no `""` value in the `args` array.


jsight:
lgtm

PR #1237, opened 2017-11-16
WINDUP-1776 --output option NPE

mrizzi:
This fixes [WINDUP-1776](https://issues.jboss.org/browse/WINDUP-1776) adding also a new test case

It has to be merged also on 4.0.x branch


jsight:
lgtm

PR #1241, opened 2017-11-22
WINDUP-1747 Bootstrap integration tests fails on CI

mrizzi:
The issue is happening on central CI so in order to better investigate it, i add the output of the whole Furnace instance **only** when it happens that input option is unrecognized due to having an empty acceptable options list because of Furnace instance being "wrong" (we're searching what's wrong).

The idea is to merge it and wait for the first failure on central CI to retrieve the information.

### DIFF
--- a/bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
+++ b/bootstrap/src/main/java/org/jboss/windup/bootstrap/Bootstrap.java
@@ -52,6 +52,7 @@ import org.jboss.windup.bootstrap.commands.windup.ListTargetTechnologiesCommand;
 import org.jboss.windup.bootstrap.commands.windup.RunWindupCommand;
 import org.jboss.windup.bootstrap.commands.windup.ServerModeCommand;
 import org.jboss.windup.bootstrap.commands.windup.UpdateRulesetsCommand;
+import org.jboss.windup.bootstrap.listener.ContainerStatusListener;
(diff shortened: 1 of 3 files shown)


mrizzi:
e3fc72f commit introduces `ContainerStatusListener` [to get notified](https://github.com/forge/furnace/blob/master/container/src/main/java/org/jboss/forge/furnace/impl/FurnaceImpl.java#L281) when Furnace is stopped w/o using the [`getStatus`](https://github.com/forge/furnace/blob/master/container/src/main/java/org/jboss/forge/furnace/impl/FurnaceImpl.java#L470) method that is based on evaluating the [`alive`](https://github.com/forge/furnace/blob/master/container/src/main/java/org/jboss/forge/furnace/impl/FurnaceImpl.java#L66) field that doesn't really ensures that Furnace is stopped.

jsight:
This seems reasonable. Have you tested it much?

jsight:
retest

mrizzi:
> This seems reasonable. Have you tested it much?

I tested it locally executing two tests `-Dtest=MissingValueOfExcludePackagesTest,SourceModeExplodedAppTest` many times and it never failed during the shutdown.

mrizzi:
retest

mrizzi:
retest

mrizzi:
retest

mrizzi:
retest

PR #1247, opened 2017-12-05
WINDUP-1791 - Fix text selection (master)

klinki:
I removed Drag & Drop code. Was there any usage for it? Can it be removed, or should I keep it there and fix it?


jsight:
I have no idea why it would have been there. I think it is right to remove it.

PR #1258, opened 2017-12-14
WINDUP-1799 - tech report heading

klinki:
Update tech bubble report to keep heading on top while scrolling table content.

Screenshot:
[](https://postimg.org/image/wgdbr419n/)


jsight:
I tried it today with a large batch of applications and the results look really nice. I only tried it in Chrome and Firefox.

PR #1267, opened 2018-02-02
WINDUP-1879 Fix XmlFile condition

mrizzi:
The `ParameterizedPatternResult` should be submitted only if the `XMLFile` condition is missing the XPath condition and not every time the file name pattern matches like it happened in [XmlFileNameValidator#isValid](https://github.com/windup/windup/compare/master...mrizzi:WINDUP-1879?expand=1#diff-885e00361e6e936a1a98a6636548f2d7L33) w/o having evaluated if there's also an XPath condition to be checked.
So the `ParameterizedPatternResult` submit has been moved to [XmlFile#registerAndSubmitResultsFor](https://github.com/windup/windup/compare/master...mrizzi:WINDUP-1879?expand=1#diff-3a9eccf2ac145f087bd0f09958ba1655R310) to be better handled.

Thanks to @jsight for the help in the analysis of this issue 👍


mrizzi:
retest

mrizzi:
retest

PR #1269, opened 2018-02-07
WINDUP-1879 Fix XmlFile condition (#1267)

mrizzi:
Back porting of #1267 
Not exactly the same code because of the switch to TP3 on master so [addVertex(null, ProjectModel.class)](https://github.com/windup/windup/compare/4.0.x...mrizzi:WINDUP-1879_4.0.x?expand=1#diff-8f24f401ef2a9a7d31bedf18ef55c35aR219) and [.addVertex(null, FileModel.class);](https://github.com/windup/windup/compare/4.0.x...mrizzi:WINDUP-1879_4.0.x?expand=1#diff-8f24f401ef2a9a7d31bedf18ef55c35aR221) have been used here.


jsight:
lgtm

PR #1270, opened 2018-02-07
Fixed a case where the graph is closed but Janus still tries to track…

jsight:
… it. This causes exceptions if the data is deleted in the web UI


jsight:
Closing as this isn't actually the right fix.

PR #1272, opened 2018-02-26
No longer reload from the graph on GraphService.refresh. This method …

jsight:
…is only intended to reload the frame itself, as that was the behavior before the TP2 port. Fixing this improves performance


jsight:
retest

windupgithubbot1:

PR #1277, opened 2018-03-14
WINDUP-1915 DiscoverEmbeddedCache*LibraryRuleProvider add cloud-readiness target

mrizzi:
1. changed all the DiscoverEmbeddedCache*LibraryRuleProvider to have the cloud-readiness target so that they will be fire only when cloud-readiness is one of the analysis' targets
2. fixed target versionRange management

### DIFF
--- a/config/api/src/main/java/org/jboss/windup/config/metadata/MetadataBuilder.java
+++ b/config/api/src/main/java/org/jboss/windup/config/metadata/MetadataBuilder.java
@@ -141,7 +141,9 @@ public class MetadataBuilder extends AbstractRulesetMetadata implements RuleProv
             {
                 builder.addTargetTechnology(new TechnologyReference(
                             technology.id(),
-                            Versions.parseVersionRange(technology.versionRange())));
+                            "".equals(technology.versionRange().trim())
+                                    ? new EmptyVersionRange()
+                                    : Versions.parseVersionRange(technology.versionRange())));
             }
         }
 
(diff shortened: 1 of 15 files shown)


config/api/src/main/java/org/jboss/windup/config/metadata/MetadataBuilder.java
```diff
@@ -141,7 +141,9 @@ public static MetadataBuilder forProvider(Class<? extends RuleProvider> implemen
             {
                 builder.addTargetTechnology(new TechnologyReference(
                             technology.id(),
-                            Versions.parseVersionRange(technology.versionRange())));
```
jsight: What is this about?

jsight:
lgtm - just one question on something that I didn't quite understand.

config/api/src/main/java/org/jboss/windup/config/metadata/MetadataBuilder.java
mrizzi: I ported the fix for handling the [`versionRange` default `""`](https://github.com/windup/windup/blob/4.0.x/config/api/src/main/java/org/jboss/windup/config/metadata/Technology.java#L25) implemented some lines above for [source technologies](https://github.com/windup/windup/blob/4.0.x/config/api/src/main/java/org/jboss/windup/config/metadata/MetadataBuilder.java#L131) to the target technologies.

jsight: Ah, that makes sense.

PR #1280, opened 2018-03-16
Ci intermittent failures

jsight:
Despite this PR appearing massive, it really only does a couple of things:
 - enable graphcontextfactory to track all graphs that it has opened and to closeAll for any open graphs
- Make GraphContextFactory clients explicitly specify whether they want graph listeners. At the moment, only one graph at a time can have these, but we also only use them on analysis graphs during analysis. So everything else should say "false".


mrizzi:
retest

PR #1282, opened 2018-04-04
WINDUPRULE-152 <where> clause working with <annotation*> tags

mrizzi:
Now `<annotation-literal>`, `<annotation-type>` and `<annotation-list>` tags work w/ `<where>` clauses

### DIFF
--- a/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/condition/JavaClass.java
+++ b/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/condition/JavaClass.java
@@ -387,6 +387,8 @@ public class JavaClass extends ParameterizedGraphCondition implements JavaClassB
             result.addAll(typeFilterPattern.getRequiredParameterNames());
         if (lineMatchPattern != null)
             result.addAll(lineMatchPattern.getRequiredParameterNames());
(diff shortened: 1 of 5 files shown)


jsight:
Do you know why the deps check fails?

jsight:
In looking through the code, it all looks correct to me. It would be nice if there was a test, though?

mrizzi:
deps check fails because w/ the latest changes [this line of code ](https://github.com/windup/windup-rulesets/blob/master/rules-reviewed/eap7/eap6/tests/data/data-hsearch/Book.java#L75) in a test class is not firing the rule anymore. 
I need to investigate.

I totally agree w/ you about adding the test.

jsight:
The issue is that parameterization doesn't know anything about or clauses. Because of this, it fails on the first pattern that doesn't match and rejects all of the matching frames.

This is the line that calls reject and ultimately rejects the frame containing these matches:
https://github.com/windup/windup/blob/master/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/condition/JavaClass.java#L192

mrizzi:
The issue has been fixed creating a Groovy rule in https://github.com/windup/windup-rulesets/pull/331

PR #1283, opened 2018-04-16
WINDUP-959: cli analyses non-java archive files

m-brophy:


windupgithubbot1:

bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/RunWindupCommand.java
```diff
@@ -541,4 +542,12 @@ private void normalizePackagePrefixes(WindupConfiguration windupConfiguration)
 
         return packages;
     }
+
+    private static boolean isJavaArchive(Path path)
```
jsight: This should use org.jboss.windup.util.ZipUtil#getZipExtensions rather than its own implementation.

m-brophy:
requested changes completed and ready to be reviewed again

PR #1284, opened 2018-04-16
WINDUP-1459: Create a link from the application list to the CSV report (if available). This allows access from the web ui as well

jsight:
https://issues.jboss.org/browse/WINDUP-1459


jsight:
retest

PR #1285, opened 2018-04-17
WINDUP-1979 Dependency report: fix name

mrizzi:
Fixed the dependency's name issue and created a test (with data) for this case.


rules-java/api/src/main/resources/reports/templates/dependency_report.ftl
```diff
@@ -76,9 +76,9 @@
                                     <dt class="trait">SHA1 hash:</dt>
                                     <dd id="${archiveName}-hash">${sha1}</dd>
                                 </#if>
-                                <#if dependencyProject.name?? && dependencyProject.name != archiveName>
+                                <#if archiveName??>
```
jsight: Now that I'm seeing the diff, I'm having some doubts. Doesn't this effectively make this whole line redundant? Line 58 already displays the archive name, right?

mrizzi: Yep, right now it's a redundant information but line 58 is the `h3` title of the panel so, in the future, it could change the content while this line takes care of showing, no matter what the title shows, the name of the dependency.
For this reason maybe is it worth keeping it?

mrizzi:
As decided during our meeting, there `Name` field has been removed since it's redundant.

jsight:
Looks like this causes a test failure?

mrizzi:
@jsight yep, you're right, thanks.
Commit https://github.com/windup/windup/pull/1285/commits/104260947f4360bfce5f55fa509714450b2a3932 fixed tests issue.

PR #1288, opened 2018-05-01
WINDUP-1957: invalid input regex breaks search in application list

m-brophy:


windupgithubbot1:

jsight:
Ah, I added you to one of the pr builders, but not the other. Fixing.

jsight:
@windupgithubbot1 retest

jsight:
I don't think this does what the user will expect. The term is still added and there is no error message displayed in the form itself (only in the browser console).

PR #1296, opened 2018-05-08
WINDUP-1967: tech report number of libraries is always 1

m-brophy:


reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/GetNumberOfLibrariesMethod.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public Object exec(@SuppressWarnings("rawtypes") List arguments) throws TemplateModelException
+    {
+        ExecutionStatistics.get().begin(NAME);
+
+        // Function arguments
+        ProjectModel projectModel = null;
+
+        // The project. May be null -> all input applications.
```
jsight: I'm pretty sure this line is wrong. Won't null just trigger a NullPointerException?

It might be better to check for null and throw a windup exception with a detailed message.

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/GetNumberOfLibrariesMethod.java
```diff
@@ -0,0 +1,87 @@
...
+            projectModel = (ProjectModel) projectArg.getWrappedObject();
+        }
+
+        Integer result = countLibrariesInModel(this.graphContext, projectModel);
+
+        ExecutionStatistics.get().end(NAME);
+        return result;
+    }
+
+    private Integer countLibrariesInModel(GraphContext graphContext, ProjectModel projectModel)
```
jsight: This looks right to me. Good job! :)

jsight:
I will test this later this afternoon, but other than one comment, this looks good to me.

jsight:
Ok, I've tested it. I just have the one comment about the comment and argument validation, but otherwise, this is working perfectly.

m-brophy:
Requested changes done

jsight:
lgtm

PR #1297, opened 2018-05-08
WINDUP-1956 Tags filter like Name filter

mrizzi:
Now  Tags filter works like Name filter.

https://issues.jboss.org/browse/WINDUP-1956


jsight:
retest

PR #1300, opened 2018-05-09
WINDUP-852: Better validation for the input path in --discoverPackages

jsight:
https://issues.jboss.org/browse/WINDUP-852


jsight:
retest

PR #1308, opened 2018-05-17
WINDUP-2015: Fixed application list sorting when there are greater than 1,000 story points

jsight:
https://issues.jboss.org/browse/WINDUP-2015


mrizzi:
retest

PR #1310, opened 2018-05-28
WINDUP-2042 Fix input path with no Java archives

mrizzi:
https://issues.jboss.org/browse/WINDUP-2042


jsight:
I tried a few different scenarios and it seems to work in all of them. Also, the test case is nice. Thanks!

PR #1311, opened 2018-06-01
WINDUP-2035: Application list: tag filter is not working as expected

jsight:
https://issues.jboss.org/browse/WINDUP-2035

### DIFF
--- a/reporting/impl/src/main/resources/reports/resources/js/windup-application-list-sorting.js
+++ b/reporting/impl/src/main/resources/reports/resources/js/windup-application-list-sorting.js
@@ -22,20 +22,32 @@ $(document).ready(function () {
             if (!filteringData || !filteringData.hasOwnProperty('tags') || filteringData.tags.length === 0) {
                 return false;
             }
+            var searchText = filterOption.data.text;
+            var regex = filterOption.data.isRegex ? new RegExp(filterOption.data.text, 'i') : null;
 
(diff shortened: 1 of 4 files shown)


reporting/impl/src/main/resources/reports/resources/js/windup-application-list-sorting.js
```diff
@@ -289,10 +300,10 @@ $(document).ready(function () {
          * @param value {string}
          * @param option {object}
          */
-        function addFilter(value, option) {
+        function addFilter(value, option, isRegex) {
             var returnValue = true;
             var filter = $.extend({}, option);
-            filter.data = value;
+            filter.data = { text: value, isRegex: isRegex };
```
mrizzi: Commenting on this line since this changes the `data` data structure so also line [`html.find('span.label').prepend(item.name + ': ' + item.data);`](https://github.com/windup/windup/pull/1311/files?utf8=%E2%9C%93&diff=split#diff-c6eb553a086747661bf08b5bcc1f241bR287) should be changed to have `item.data.text` otherwise, once a tag is selected, is not showed in the right way in the `Active filters` top bar

jsight: Fixed! Thanks for catching that. In the process, I also noticed some text inconsistencies, so I've added that to the PR as well.

PR #1312, opened 2018-06-05
WINDUPRULE-331: Also made it possible to use iteration filtering in t…

jsight:
…he way that a user would expect


mrizzi:
retest

jsight:
Moving back to WIP... looks like there is a intermittent test failure, so I should probably think about the annotation test a little more.

jsight:
retest

jsight:
@mrizzi - I think this is ready to review again.

PR #1314, opened 2018-06-08
change techReport-boxes font-sizes to be more consistent with the other reports

jeichler:
The tech report boxes use a different font size (12pt) than for instance the Issues page (12px).
I kept the headers for the boxes at 12pt.

This led to too early line breaks in the tech boxes (as browser window is maximised with a 15" screen), since the 12pt got rendered on my current test box as 16px - plus it's not really consistent.

While being there I did very minor changes to the font sizes of the JavaEE /Embedded text (they now match the size of e.g. "Migration Mandatory"on the issues page) and the text for the tiers (Connect, Sustain, etc. - they now match the size in the bubble map) - it's not perfect, but it's more consistent now.

Of course, when minimising the browser window, this will happen eventually, but higher consistency plus lesser line breaks on "standard screens" is probably a good thing to have.

before:
screen shot 2018-06-08 at 20 50 27

Note here the cassandra client or the ocpsoft logging utils where you can see the issue.

after:
screen shot 2018-06-09 at 00 49 01

I tested the changes on Mac (Safari, Chrome, Firefox), Linux (Firefox, Chrome) and Windows (Edge, Chrome, Firefox).

### DIFF
--- a/reporting/impl/src/main/resources/reports/templates/techReport-boxes.ftl
+++ b/reporting/impl/src/main/resources/reports/templates/techReport-boxes.ftl
@@ -52,7 +52,7 @@
         tr.rowSectors > td  > div.box { padding: 8pt 18pt 8pt 12pt; margin-bottom: 10pt; }
         tr.rowSectors > td  > div.box > h4 { font-size: 12pt; font-weight: bold; }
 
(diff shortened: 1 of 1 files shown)


windupgithubbot1:

windupgithubbot1:

jsight:
@windupgithubbot1 test this please

PR #1317, opened 2018-06-25
WINDUP-2059 Execute with jboss-release and gpg-sign profiles

mrizzi:


mrizzi:
Already discussed last week with @jsight so going to merge it directly.

PR #1321, opened 2018-06-27
WINDUP-1998: Upgrade to latest forge and furnace releases

jsight:


mrizzi:
Thanks @jsight for starting this task.
This work will be done in https://issues.jboss.org/browse/WINDUP-2171 and https://github.com/windup/windup/pull/1332

PR #1322, opened 2018-07-02
Fixed a case where the fileservice doesn't behave well if a filemodel…

jsight:
… entry was precreated.

This really only occurs in odd ege cases. For example, if you were to analyze a source directory which contained a user rules directory. The configuration would create FileModels
for the rules directory, which would later be reused, but their parents wouldn't be setup correctly.


mrizzi:
retest

mrizzi:
retest
since https://github.com/windup/windup/pull/1324 has been merged

PR #1330, opened 2018-08-23
WINDUP-2018 Enable tattletale by default

mrizzi:
- Tattletale report generation enabled by default without any need to provide option
- deprecated `--enableTattletale` option with the following console warning:
```
WARNING: (DEPRECATED) --enableTattletale option is not necessary anymore since Tattletale report generation is enabled by default. Use only --disableTattletale option if you want to disable it.
```
- added `--disableTattletale` option
- check before executing analysis' execution that not both options (`--enableTattletale` and `--disableTattletale`) are provided. If that's the case the executions stops with the following message:
```
ERROR: Do NOT use --enableTattletale and --disableTattletale options together. Tattletale report generation is enabled by default so --enableTattletale option should be removed.
```
- created new tests for different use cases (no option, only `--enableTattletale`, only `--disableTattletale`, both options)
- this change works also in the Web UI without any needed change
- Tattletale report: add the standard header (with tooltip, application name and paddings) we have in all the other reports (and removing the standard border the browser adds to iframes)
screenshot from 2018-08-23 19-23-45


mrizzi:
Commit e44fc8a has the changes needed for implementing the requirements from [this comment on JIRA](https://issues.jboss.org/browse/WINDUP-2018?focusedCommentId=13623640&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13623640) on.

rules-tattletale/addon/src/main/java/org/jboss/windup/rules/apps/tattletale/TattletaleRuleProvider.java
```diff
@@ -58,8 +60,13 @@ public Configuration getConfiguration(RuleLoaderContext ruleLoaderContext)
         @Override
         public void perform(GraphRewrite event, EvaluationContext context)
         {
-            Boolean generateReport = (Boolean) event.getGraphContext().getOptionMap().get(EnableTattletaleReportOption.NAME);
-            if (generateReport == null || !generateReport)
+            Boolean enableReport = (Boolean) event.getGraphContext().getOptionMap().get(EnableTattletaleReportOption.NAME);
+            Boolean disableReport = (Boolean) event.getGraphContext().getOptionMap().get(DisableTattletaleReportOption.NAME);
+            Collection<String> targets = (Collection<String>) event.getGraphContext().getOptionMap().get(TargetOption.NAME);
+            boolean eapTarget = targets.stream().anyMatch(target -> target.startsWith("eap"));
+            if (eapTarget && disableReport != null && disableReport && (enableReport == null || !enableReport))
```
jeichler: for a tiny bit more readability I would use Boolean.valueOf(disableReport) here and for the enableReport logic as well

reporting/impl/src/main/resources/reports/templates/embedded.ftl
```diff
@@ -25,7 +25,24 @@
     </div>
     <!-- / Navbar -->
 
-    <iframe style="width:100%; height: 100%; position:absolute;" src="${reportModel.reportProperties.embeddedUrl}"></iframe>
+    <div class="container-fluid" role="main">
```
jonathanvila: Is this piece of code related with the change or boy-scout refactoring ?

rules-tattletale/addon/src/main/java/org/jboss/windup/rules/apps/tattletale/TattletaleRuleProvider.java
```diff
@@ -58,8 +61,13 @@ public Configuration getConfiguration(RuleLoaderContext ruleLoaderContext)
         @Override
         public void perform(GraphRewrite event, EvaluationContext context)
         {
-            Boolean generateReport = (Boolean) event.getGraphContext().getOptionMap().get(EnableTattletaleReportOption.NAME);
-            if (generateReport == null || !generateReport)
+            Boolean enableReport = (Boolean) event.getGraphContext().getOptionMap().get(EnableTattletaleReportOption.NAME);
+            Boolean disableReport = (Boolean) event.getGraphContext().getOptionMap().get(DisableTattletaleReportOption.NAME);
+            Collection<String> targets = (Collection<String>) event.getGraphContext().getOptionMap().get(TargetOption.NAME);
+            boolean eapTarget = targets != null && targets.stream().anyMatch(target -> target.startsWith("eap"));
+            if (eapTarget && BooleanUtils.isTrue(disableReport) && BooleanUtils.isNotTrue(enableReport))
```
jonathanvila: As this is not a straight forward condition, what do you think on extracting a method in charge of decide if the report is enabled ?

jonathanvila:
I have to say that from the developer point of view is a bit criptic having enableOption and disableOption at the same time, and classes with a name that includes the enabling status.
I understand that is for backward compatibility, but having something proxing the external configuration to a class that only considers an option and has its enabled status could be a bit more clear.

reporting/impl/src/main/resources/reports/templates/embedded.ftl
mrizzi: Yes it is.
It's the last entry (in my first comment above) in the list of changes required: 
_Tattletale report: add the standard header (with tooltip, application name and paddings) we have in all the other reports (and removing the standard border the browser adds to iframes)_

rules-tattletale/addon/src/main/java/org/jboss/windup/rules/apps/tattletale/DisableTattletaleReportOption.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.jboss.windup.config.AbstractConfigurationOption;
+import org.jboss.windup.config.InputType;
+import org.jboss.windup.config.ValidationResult;
+
+/**
+ *
+ * Indicates that the Tattletale report should NOT be generated.
+ *
+ * @author <a href="mailto:marcorizzi82@gmail.com">Marco Rizzi</a>
```
jonathanvila: Do we have to add our email in the classes we create ?

mrizzi: I think it's good to have the author in a class.
I don't think we should consider mandatory to have the email there as well.

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/DisableTattletaleParameterTest.java
```diff
@@ -0,0 +1,53 @@
+package org.jboss.windup.tests.bootstrap.migrate;
+
+import org.jboss.windup.bootstrap.Bootstrap;
+import org.jboss.windup.rules.apps.tattletale.TattletaleRuleProvider;
```
jonathanvila: it seems this import is not used

mrizzi: good catch, thanks.
Also [TattletaleEnablementByDefaultTest.java](https://github.com/windup/windup/pull/1330/files#diff-aedaa08448dcdd9e9b21f04ae02e2794) seems to have some

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/DisableTattletaleParameterTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class DisableTattletaleParameterTest extends AbstractBootstrapTestWithRules {
+    @Rule
+    public final TemporaryFolder tmp = new TemporaryFolder();
+
+    @Rule
+    public final TemporaryFolder tmpAddonDir = new TemporaryFolder();
+
+    @Test
+    public void disableTattletaleParameterWithEapTarget() {
```
jonathanvila: are we following any approach for the name of the tests ? 
https://dzone.com/articles/7-popular-unit-test-naming

mrizzi: I think our approach looks like `4. Feature to be tested`

rules-tattletale/addon/src/main/java/org/jboss/windup/rules/apps/tattletale/TattletaleRuleProvider.java
```diff
@@ -145,5 +147,18 @@ private void createReportModel(GraphContext context, FileModel input, String rep
             ReportService reportService = new ReportService(context);
             reportService.setUniqueFilename(applicationReportModel, "tattletale" + "_" + inputProjectModel.getName(), "html");
         }
+
+        private boolean isTattleTaleReportGenerationEnabled(GraphRewrite event)
+        {
+            Boolean enableReport = (Boolean) event.getGraphContext().getOptionMap().get(EnableTattletaleReportOption.NAME);
```
jonathanvila: We could convert this to :
`
            boolean enableReport = (boolean) event.getGraphContext().getOptionMap().getOrDefault(EnableTattletaleReportOption.NAME, Boolean.FALSE);`

rules-tattletale/addon/src/main/java/org/jboss/windup/rules/apps/tattletale/TattletaleRuleProvider.java
```diff
@@ -145,5 +147,18 @@ private void createReportModel(GraphContext context, FileModel input, String rep
...
+            Boolean disableReport = (Boolean) event.getGraphContext().getOptionMap().get(DisableTattletaleReportOption.NAME);
+            Collection<String> targets = (Collection<String>) event.getGraphContext().getOptionMap().get(TargetOption.NAME);
+            boolean eapTarget = targets != null && targets.stream().anyMatch(target -> target.startsWith("eap"));
+            if (eapTarget && BooleanUtils.isTrue(disableReport) && BooleanUtils.isNotTrue(enableReport))
+                return false;
+            else if (!eapTarget && BooleanUtils.isNotTrue(enableReport))
+                return false;
+            return true;
```
jonathanvila: At this point what I see is that we cover this cases :
1. eap = true , disable = true, enable = true ----> false
2. eap = false, enable = false ---> false

But we are covering these 2 cases as true :
4. eap = false, enable = true ---> true
5. eap = true, disable = false, enable = false ---> true
Are we totally sure with case #5 ?

mrizzi: Just to keep trace, we talked about the use cases in IRC and figured out that these options are `true` if provided in command line and `false` otherwise.

jonathanvila: +1

rules-tattletale/addon/src/main/java/org/jboss/windup/rules/apps/tattletale/TattletaleRuleProvider.java
mrizzi: I'm afraid this would lead to a NPE in case there's no such a key due to the unboxing done on `null`

mrizzi: Ah sorry, you proposed the default version even if i don't think we should have the options default here.
Would you prefer having the `boolean` instead of `BooleanUtils` for readability?

jonathanvila: well, a pure boolean is easier to read, and in fact if we dont have the value in the map we consider them as false, is what BooleanUtils is doing 
`public static boolean isTrue(Boolean bool) {
        if (bool == null) {
            return false;
        } else {
            return bool;
        }
    }`

PR #1334, opened 2018-09-26
 WINDUP-2137 Dependency graph 

mrizzi:
https://issues.jboss.org/browse/WINDUP-2137

Thanks @jeichler for providing the first implementation of this feature.


PhilipCattanach:
Hi Marco, 
Overall very good. There are just a couple of minor issues that I have noticed.
When the AdministracionEfectivo.ear application is analyzed the Dependency Graph shows the jar files. If you interact with the graph spreading the .jar nodes across the page, then click on the individual .jar nodes in turn, sometimes they reposition themselves (rebound) more centrally. 
One in particular (the -seguridad- jar) the hover-over name doesn't appear, not can you click on the node to get the file name to appear on the left hand side of the screen.

jonathanvila:
I have tested the feature against 2 applications, one very simple and another more complex.
With the second one I see something very weird as I dont have scroll bars so I'm missing bottom part of the graph, and also creates a funny representation with icons on the edges.....
image

jonathanvila:
It creates a very "weird" behaviour while creating the nodes on the screen , or touching one of them.
Videos : 
https://drive.google.com/open?id=1biSrJ4q7BR2MSNP1dWXa9Bmhn3v3RPTj
https://drive.google.com/open?id=1jzhS6MeFPIbuuHhiw-GHBkQ3op3zY-FB

jeichler:
in your videos, @jonathanvila it also seems that some nodes are not connected at all, which should never be the case for the single app report as it is all packaged within the very same app, so they should have at least one connection to the input file / root model, right?

jeichler:
on the report itself, the info icon which usually displays on mouse over a short explanation, doesn't have any text to render.

jeichler:
the direction of the relation (source / target) seems sometimes to be wrong (e.g. 3rd party jars have dotted lines in one single app report, and non-dotted in antoher single app report).

jonathanvila:
@jeichler My thoughts are : 
* from an user perspective I only see a graph of icons, no texts... In a complex application which is the real value we want to give with this report ?
* we are using a library intended to graph services in a K8s architecture.... Are we sure this is what we need ?
* this "funny" movement , is it needed ?
* we dont have features like : export ( in a text language ) , dig, panel with artifact info, etc...
* is it mandatory to use this library to create the graph ?

mrizzi:
@PhilipCattanach, @jonathanvila and @jeichler thanks for testing this PR.
I try to sum up the evidences about the dependency graph:

1. libraries "alone" (i.e. nodes not connected to any other node): this is due to the analyzed application that it's not web application but it's a zip file with inside some folders with a lot of JARs so it's right for RHAMT creating node not connected since there's no web application they refer to
2. dotted vs solid lines: i'm going to push a commit to remove the dotted lines as requested
3. war application with a line to a dependency (`terracotta-toolkit-1.6-runtime-5.0.0.jar`) and then further dependencies connected to the latter dependency rather to the war application: there's no bug, it's working in the right way since the `terracotta-toolkit-1.6-runtime-5.0.0.jar` is an archive containing other jar archives
4. the info icon on the title bar is showing the tooltip
5. the case of huge number of circles will be solved in another issue ([WINDUP-2183](https://issues.jboss.org/browse/WINDUP-2183))

@jeichler: so, from the functional point of view, do you see any further pending issue to be solved in this PR?

jeichler:
@mrizzi thanks.

point 4 is somehow working but not correctly. The wrong css is used on that report for displaying the tooltip (and probably the wrong js too since it has the same delay as the mouse over text for the nodes on the dependency graph nodes). So this should be fixed.

point 3: It is true, I am / was blind! Thanks for looking into it :)

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/dependencyreport/CreateDependencyReportRuleProvider.java
```diff
@@ -150,7 +149,7 @@ private void createGlobalReport(GraphContext context, WindupConfigurationModel c
             addAll(context, reportModel, traversal, sha1ToGroup);
         }
 
-        reportModel.setDisplayInGlobalApplicationIndex(true);
+        reportModel.setDisplayInGlobalApplicationIndex(Boolean.TRUE);
```
jonathanvila: what's the difference ? the autoboxing would directly convert to Boolean, right ?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/freemarker/dto/DependencyGraphItem.java
```diff
@@ -0,0 +1,86 @@
...
+   private final Map<String, String> metadata;
+
+   public DependencyGraphItem(DependencyReportDependencyGroupModel dependencyReportDependencyGroupModel)
+   {
+      this(dependencyReportDependencyGroupModel.getCanonicalProject());
+   }
+
+   public DependencyGraphItem(ProjectModel projectModel)
+   {
+      this.metadata = new HashMap<>(1);
```
jonathanvila: why do we need to use "this." ?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateDependencyGraphReportRuleProvider.java
```diff
@@ -0,0 +1,96 @@
...
+
+   private class CreateDependencyGraphReportOperation extends GraphOperation
+   {
+
+      @Override
+      public void perform(GraphRewrite event, EvaluationContext context)
+      {
+         List<FileModel> inputPaths = WindupConfigurationService.getConfigurationModel(event.getGraphContext())
+                  .getInputPaths();
+         if (Iterables.size(inputPaths) > 1)
```
jonathanvila: Why this Iterables.size instead of a simple inputPaths.size() ? Both choices are not "null protected"

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/dependencyreport/CreateDependencyReportRuleProvider.java
jsight: lol, I noticed that and wondered the same thing.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/freemarker/dto/DependencyGraphItem.java
jsight: A former developer on the project used to complain about member variable references that didn't use "this.". I have a feeling that it became a habit since then for some of us.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateDependencyGraphReportRuleProvider.java
jsight: +1 - I suspect copy'pasta as some of these used to be Iterables instead of Lists.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateDependencyGraphDataRuleProvider.java
```diff
@@ -0,0 +1,158 @@
...
+
+         // in case of shared libraries ("virtual" project) we don not
+         // generate the dependency graph
+         if (application != null && ProjectModel.TYPE_VIRTUAL.equals(application.getProjectType()))
+            return;
+
+         Path dataDirectory = reportService.getReportDataDirectory();
+         List<DependencyReportDependencyGroupModel> dependencyReportDependencyGroupModels = dependenciesReportModel
+                  .getArchiveGroups();
+         Map<String, DependencyGraphItem> items = new HashMap<>(dependencyReportDependencyGroupModels.size() + 1);
```
m-brophy: What's the +1 for?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateDependencyGraphReportRuleProvider.java
```diff
@@ -0,0 +1,96 @@
...
+                  .filter(inputPath -> !ProjectModel.TYPE_VIRTUAL.equals(inputPath.getProjectModel().getProjectType()))
+                  .forEach(inputPath -> createSingleAppDependencyGraphReport(event.getGraphContext(),
+                           inputPath.getProjectModel()));
+      }
+
+      private ApplicationReportModel createAppDependencyGraphReport(GraphContext context)
+      {
+         ApplicationReportService applicationReportService = new ApplicationReportService(context);
+         ApplicationReportModel report = applicationReportService.create();
+         report.setReportPriority(104);
```
m-brophy: Perhaps the '104' could be a constant?

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/dependencyreport/CreateDependencyReportRuleProvider.java
mrizzi: ....hhhmmm....other than this change, i don't know why this class has been changed since it's not involved in implementing this feature and i'm quite sure i didn't change it :thinking: 

I figured out what happened: it was part of the initial commit from Janine [CreateDependencyReportRuleProvider](https://github.com/jeichler/windup/blob/dependencies-graph/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/dependencyreport/CreateDependencyReportRuleProvider.java#L152) so i didn't see it so thanks for highlighting this.
We're not supposed to avoid the autoboxing so this kind of changes are not expected to be done.

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateDependencyGraphReportRuleProvider.java
mrizzi: Good catch: i'm going to change it.
And yes, as Jess highlighted, some of these are defined as `Iterable`s

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateDependencyGraphDataRuleProvider.java
mrizzi: It's for the analyzed application itself that is not part of the its own `List<DependencyReportDependencyGroupModel> dependencyReportDependencyGroupModels`

reporting/impl/src/main/resources/reports/templates/dependency_graph.ftl
```diff
@@ -0,0 +1,193 @@
...
+        <#--<title>-->
+            <#--<#if reportModel.projectModel??>-->
+                <#--${reportModel.projectModel.name} --->
+            <#--</#if>-->
+            <#--${reportModel.reportName}-->
+        <#--</title>-->
+        <meta charset="utf-8">
+        <title>Application Dependency Graph</title>
+        <meta name="viewport" content="width=device-width">
+        <style>
```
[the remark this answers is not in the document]
mrizzi: That's true, we can move it, thanks :+1:

rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/reporting/rules/CreateDependencyGraphReportRuleProvider.java
mrizzi: It could but in this case it's a "one time" usage just to set this value and there's no further need to reference this value somewhere else in the code so i think it's anyway fine to have it define in this way.

mrizzi:
@jeichler it took me some time to understand but i solved the tooltip issue in the latest commit :wink:

mrizzi:
retest

PR #1339, opened 2018-10-23
WINDUP-2237 Fix the exception during graph generation

mrizzi:
https://issues.jboss.org/browse/WINDUP-2237
- fixes the exception during the dependency graph report generation
- adds a node in the graph for the exploded application (considered as a WAR archive)
- adds the test for the dependency graph generation for the  `--explodedApp` use case

### DIFF
--- a/reporting/impl/src/main/java/org/jboss/windup/reporting/freemarker/Sha1HexMethod.java
+++ b/reporting/impl/src/main/java/org/jboss/windup/reporting/freemarker/Sha1HexMethod.java
@@ -1,0 +1,50 @@
+package org.jboss.windup.reporting.freemarker;
+
+import java.util.List;
+
+import freemarker.ext.beans.StringModel;
+import org.apache.commons.codec.digest.DigestUtils;
+import org.jboss.windup.config.GraphRewrite;
+
+import freemarker.template.TemplateModelException;
+import org.jboss.windup.graph.model.resource.FileModel;
+
+/**
+ * Returns the SHA1 Hex hash for the provided {@link FileModel}.<br/>
+ * 
+ * sha1Hex(FileModel):String
+ *
+ */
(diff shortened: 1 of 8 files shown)


jsight:
Can't you use the "--explodedApp" switch?

mrizzi:
The issue happens when the `--explodedApp` switch is used since there's no parent archive model for the JAR libraries found in the exploded application and so there wre no lines to represent the relations in the dependency graph.

mrizzi:
This is how a "traditional" WAR application's dependency graph looks like:

screenshot from 2018-10-23 15-50-44

And the next one is the dependency graph created (using code from commit https://github.com/windup/windup/pull/1339/commits/f944cd7ffc37e119de56b5261f8b6bd06239ac9d) during the analysis of the same application above but unzipped and so using the `--explodedApp` option:

screenshot from 2018-10-23 15-50-37

jsight:
Oh, I see what you mean. Hmm.... I would have expected there to be dependency relationships anyway. Maybe the model is a little strange in this case, though, since the parent is no longer an archive.

mrizzi:
Exactly, you got the point: in a standard analysis, each library archive has the parent set to the analyzed archive during the unzipping.
In the exploded application use case, there's no unzipping and so the library archives have no parent that, in terms of dependency graph, means having no line for relations.

mrizzi:
Just to further describe the issue, after the first commit (https://github.com/windup/windup/pull/1339/commits/276310de6b8e3aafa9aa7f956ab9ef7589a631e6 -  to fix the exception reported in [WINDUP-2237](https://issues.jboss.org/browse/WINDUP-2237)) the dependency graph for exploded `spring4-jms-weblogic-1.0-SNAPSHOT` application looked like this - without any line:
screenshot from 2018-10-23 16-36-38

PR #1342, opened 2018-12-05
WINDUP-2258 Enhanced TarUtil to exclude some files from archiving

mrizzi:
- new environmental property `FULL_TAR_ARCHIVE` to force the creation of a Tar file with everything found inside the input path directory
- the ability to provide a list of paths' prefixes to be excluded from the tar archive

### DIFF
--- a/utils/src/main/java/org/jboss/windup/util/TarUtil.java
+++ b/utils/src/main/java/org/jboss/windup/util/TarUtil.java
@@ -12,6 +12,7 @@ import java.nio.file.Path;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.logging.Logger;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.kamranzafar.jtar.TarEntry;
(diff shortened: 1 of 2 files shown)


utils/src/main/java/org/jboss/windup/util/TarUtil.java
```diff
@@ -19,9 +23,19 @@
...
+
     public static void tarDirectory(Path outputFile, Path inputDirectory) throws IOException
+    {
+        tarDirectory(outputFile, inputDirectory, Collections.emptyList());
+    }
+
+    public static void tarDirectory(Path outputFile, Path inputDirectory, List<String> pathPrefixesToExclude) throws IOException
     {
         System.out.println("Creating archive at: " + outputFile);
+        boolean fullTarArchive = Boolean.valueOf(System.getenv(FULL_TAR_ARCHIVE));
```
jsight: I'm sure this works, but the placement of the code feels really strange to me. Why should a generic looking TarUtil know about a boolean FULL_TAR_ARCHIVE property and ignore one of its parameters based upon this outside value?

It feels like this logic should be in the caller.

Also, I'm assuming that you don't mean to leave the System.out.println statements in. :)

mrizzi: The basic idea was about having:
- the code to call the method passing the list `pathPrefixesToExclude`
- the user being able, at runtime, to force the TarUtil to ignore this list for generating archives with everything inside (kind of debug purpose)

Maybe it won't be so useful but this was the idea for implementing in this way.

Good point about `System.out`: if found useful the `Creating archive at:` one and i added another :sweat_smile:

jsight: I don't disagree about having the variable, it just feels like it would be better done by the caller instead of the utility function.

mrizzi:
retest

PR #1343, opened 2018-12-05
WINDUP-2179: summary and detail tech report count discrepancy

m-brophy:
results map created for techbox report was skipping some of the results, finding that the techstat was not included in the application. This fix amends how that section of the code determines if a stat is to be included in an application's techbox report.

### DIFF
--- a/reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
+++ b/reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
@@ -167,9 +167,9 @@ public class TechReportService
     /**
      * Prepares a precomputed matrix - map of maps of maps: rowTag -> boxTag -> project -> placement label -> TechUsageStatSum.
      *
-     * @param onlyForApplication Sum the statistics only for this project.
+     * @param appBeingSummarized Sum the statistics only for this project.
      */
-    public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
+    public TechStatsMatrix getTechStatsMap(ProjectModel appBeingSummarized)
     {
 
         Map<String, Map<String, Map<Long, Map<String, TechUsageStatSum>>>> map = new HashMap<>();
@@ -177,11 +177,11 @@ public class TechReportService
         final Iterable<TechnologyUsageStatisticsModel> statModels = graphContext.service(TechnologyUsageStatisticsModel.class).findAll();
         for (TechnologyUsageStatisticsModel stat : statModels)
         {
-            // A shortcut.
-            if (onlyForApplication != null &&
-                    (!stat.getProjectModel().getApplications().contains(onlyForApplication)
+            //If a stat doesn't apply to this project we move on to the next stat without further processing
+            if (appBeingSummarized != null &&
+                    (!stat.getProjectModel().getApplications().contains(appBeingSummarized)
                     || (stat.getProjectModel().getApplications().size() > 1
-                            && !onlyForApplication.getName().toLowerCase().contains("shared"))))
+                            && !appBeingSummarized.getName().toLowerCase().contains("shared"))))
             {
                 LOG.fine("\t\tThis stat is not for this project, skipping.");
                 continue;
@@ -209,7 +209,7 @@ public class TechReportService
             mergeToTheRightCell(map, "", placement.box.getName(), 0L, "", stat, true);
 
             List<Long> appsToCountTowards;
-            if (onlyForApplication == null)
+            if (appBeingSummarized == null)
             {
                 appsToCountTowards = StreamSupport.stream(stat.getProjectModel().getApplications().spliterator(), false)
                             .map(ProjectModel::getElement)
@@ -219,7 +219,7 @@ public class TechReportService
             }
             else
             {
-                appsToCountTowards = Collections.singletonList(onlyForApplication.getId());
+                appsToCountTowards = Collections.singletonList(appBeingSummarized.getId());
             }
 
             boolean isSharedApp = appsToCountTowards.size() > 1;
--- a/reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
+++ b/reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
@@ -180,7 +180,7 @@ public class TechReportService
         for (TechnologyUsageStatisticsModel stat : statModels)
         {
             // A shortcut.
-            if (applicationProjects != null && !applicationProjects.contains(stat.getProjectModel()))
+            if (onlyForApplication != null && !stat.getProjectModel().getApplications().contains(onlyForApplication))
             {
                 LOG.fine("\t\tThis stat is not for this project, skipping.");
                 continue;
--- a/reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
+++ b/reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
@@ -178,13 +178,16 @@ public class TechReportService
         for (TechnologyUsageStatisticsModel stat : statModels)
         {
             //If a stat doesn't apply to this project we move on to the next stat without further processing
-            if (appBeingSummarized != null &&
-                    (!stat.getProjectModel().getApplications().contains(appBeingSummarized)
-                    || (stat.getProjectModel().getApplications().size() > 1
-                            && !appBeingSummarized.getName().toLowerCase().contains("shared"))))
+            if (appBeingSummarized != null)
             {
-                LOG.fine("\t\tThis stat is not for this project, skipping.");
-                continue;
+                boolean appIsContainedInStatProject = stat.getProjectModel().getApplications().contains(appBeingSummarized);
+                boolean appIsASharedLibrary = appBeingSummarized.getName().toLowerCase().contains("shared");
+                boolean statAppliesToSeveralApps = stat.getProjectModel().getApplications().size() > 1;
+                if(!appIsContainedInStatProject || (statAppliesToSeveralApps && !appIsASharedLibrary))
+                {
+                    LOG.fine("\t\tThis stat is not for this project, skipping.");
+                    continue;
+                }
             }
 
             final Set<String> placementTags = TechReportService.getPlacementTags(graphContext, stat.getTags());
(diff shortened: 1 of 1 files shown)


mrizzi:
retest

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -180,7 +180,7 @@ public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
         for (TechnologyUsageStatisticsModel stat : statModels)
         {
             // A shortcut.
-            if (applicationProjects != null && !applicationProjects.contains(stat.getProjectModel()))
+            if (onlyForApplication != null && !stat.getProjectModel().getApplications().contains(onlyForApplication))
```
mrizzi: The change seems fine and i think it could be good to delete the [`Set<ProjectModel> applicationProjects`](https://github.com/windup/windup/pull/1343/files?utf8=%E2%9C%93&diff=split#diff-6670edb730de259392f0270dbec38b2eR174) declaration above (sorry i can not comment on that line directly because it's "outside" of the changes in this commit)

m-brophy:
Change made!

m-brophy:
@mrizzi I think TechReportServiceTest has been failing for a couple of builds now so I need to look into that

mrizzi:
retest

m-brophy:
@mrizzi I have made changes which hopefully solve the problems with the previous PR submission

m-brophy:
retest

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -171,16 +171,17 @@ private static TagModel getNonPlaceParent(TagGraphService tagService, TagModel t
...
     {
-        Set<ProjectModel> applicationProjects = onlyForApplication == null ? null
-                    : new ProjectModelTraversal(onlyForApplication).getAllProjects(true);
 
         Map<String, Map<String, Map<Long, Map<String, TechUsageStatSum>>>> map = new HashMap<>();
 
         final Iterable<TechnologyUsageStatisticsModel> statModels = graphContext.service(TechnologyUsageStatisticsModel.class).findAll();
         for (TechnologyUsageStatisticsModel stat : statModels)
         {
             // A shortcut.
```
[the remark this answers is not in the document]
m-brophy: Comment has been expanded to be more informative

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -171,16 +171,17 @@ private static TagModel getNonPlaceParent(TagGraphService tagService, TagModel t
      */
     public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
```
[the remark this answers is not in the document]
m-brophy: I've renamed that variable to "appBeingSummarized" which is a bit more descriptive of what's going on

m-brophy:
retest

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -167,20 +167,21 @@ private static TagModel getNonPlaceParent(TagGraphService tagService, TagModel t
...
-     * @param onlyForApplication Sum the statistics only for this project.
+     * @param appBeingSummarized Sum the statistics only for this project.
      */
-    public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
+    public TechStatsMatrix getTechStatsMap(ProjectModel appBeingSummarized)
     {
-        Set<ProjectModel> applicationProjects = onlyForApplication == null ? null
-                    : new ProjectModelTraversal(onlyForApplication).getAllProjects(true);
 
         Map<String, Map<String, Map<Long, Map<String, TechUsageStatSum>>>> map = new HashMap<>();
```
jonathanvila: What do you think about refactoring this conditional ? It took me a while to understand what is filtering.

boolean appIsContainedInStatProject = stat.getProjectModel().getApplications().contains(onlyForApplication);
boolean appIsASharedApp = onlyForApplication.getName().toLowerCase().contains("shared");
boolean projectHasSeveralApps = stat.getProjectModel().getApplications().size() > 1;

if (!appIsContainedInStatProject || (pojectHasSeverlApps && !appIsASharedApp) {

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -208,7 +209,7 @@ public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
             mergeToTheRightCell(map, "", placement.box.getName(), 0L, "", stat, true);
 
             List<Long> appsToCountTowards;
-            if (onlyForApplication == null)
```
jonathanvila: This list could be retrieved out of the loop therefore not retrieving the same list several times ?

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -218,11 +219,23 @@ public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
...
             }
 
+            boolean isSharedApp = appsToCountTowards.size() > 1;
+            List<ProjectModel> apps = stat.getProjectModel().getApplications();
+
             for (Long appToCountTowards : appsToCountTowards)
             {
+                if(isSharedApp)
+                {
+                    List<ProjectModel>sharedApps = apps.stream().filter(app -> app.getName().toLowerCase().contains("shared")).collect(toList());
```
jonathanvila: Could we get this list out of the loop ? In order to not retrieve the same list several times.

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -218,11 +219,23 @@ public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
             }
             else
             {
-                appsToCountTowards = Collections.singletonList(onlyForApplication.getId());
+                appsToCountTowards = Collections.singletonList(appBeingSummarized.getId());
             }
 
+            boolean isSharedApp = appsToCountTowards.size() > 1;
```
jonathanvila: Why here the concept of sharedApp doesnt have "shared" in the name of the app and only if the list of apps to be considered is greater than 1 ?

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
```diff
@@ -218,11 +219,23 @@ public TechStatsMatrix getTechStatsMap(ProjectModel onlyForApplication)
...
+            List<ProjectModel> apps = stat.getProjectModel().getApplications();
+
             for (Long appToCountTowards : appsToCountTowards)
             {
+                if(isSharedApp)
+                {
+                    List<ProjectModel>sharedApps = apps.stream().filter(app -> app.getName().toLowerCase().contains("shared")).collect(toList());
+                    if(!sharedApps.stream().anyMatch(app -> app.getElement().id().equals(appToCountTowards)))
```
jonathanvila: Could we remove getting the sharedApps list and instead do this filter ? What do you think ?

if ( apps.stream().anyMatch(app -> app.getName().toLowerCase().contains("shared") && app.getElement().id().equals(appToCountTowards) )

jonathanvila:
I haven't seen any test modification for these changes. Are the use cases in the current existing Test covering these changes ?

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
m-brophy: > I haven't seen any test modification for these changes. Are the use cases in the current existing Test covering these changes ?

The original changes caused the tests to fail - these changes fix that

m-brophy:
> I haven't seen any test modification for these changes. Are the use cases in the current existing Test covering these changes ?
 
The original changes caused the tests to fail - these changes fix that

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
m-brophy: There's a separate stat model in the list for a technology, for each app it applies to, and there's also a separate stat model if the library is shared ie, if there are 2 apps in an analysis sharing a library, there will be 3 stat models, and the shared library one applies to all 3 apps. Only the one which is for the shared stat applies to more than 1 app.

If it's the shared stat we only want the count to apply to the shared library app, not the actual analysed apps because their counts are covered by the other stat models for this technology. So we use a count of apps within the stat greater than 1 to flag that this is the shared library stat. When we know that, we only add the count from the stat to the appropriate places in the techstatsmap for the app with 'shared' in its name, not the other ones.

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
m-brophy: I tried this but it is subtly different to the original and doesn't produce the correct results so I've left it with the sharedApps list retrieval moved outside of the loop and the original stream/filtering code

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
jonathanvila: +1

reporting/impl/src/main/java/org/jboss/windup/reporting/rules/generation/techreport/TechReportService.java
jonathanvila: +1

carlosthe19916:
I've tested this PR and it solves both [WINDUP-2179]( https://issues.jboss.org/browse/WINDUP-2179 ) and [WINDUP-2275](https://issues.jboss.org/browse/WINDUP-2275) nicely... There were some changes suggested by @jonathanvila and @mrizzi, but @m-brophy has already made changes to solve their concerns... As far as I'm concern, I think this PR should be merged because it works well.

mrizzi:
retest

PR #1346, opened 2019-02-06
WINDUP-2032: sort tech bubble report by sector header

m-brophy:


m-brophy:
retest

PhilipCattanach:
Problem identified with the column sorting, but the Tier sorting looks fine. Details attached to the Jira.

reporting/impl/src/main/resources/reports/templates/techReport-punchCard.ftl
```diff
@@ -306,15 +319,87 @@
...
+            }
+
+            $.each(rows, function(index, row) {
+                $('.technologiesPunchCard').children('tbody').append(row);
+            });
+        }
+
         $().ready(function () {
             $(".headersGroup .sector").click(function (event) {
                 var td = event.target.parentNode;
```
carlosthe19916: I noticed that every column of the header has a div inside, but because the div is shorter than the column(td) there is a problem when I click on the space between the div and the td (I'm attaching an image to explain this better)...

The problem is that whenever I click in the yellow part of the image the sort column is always the first one and not the column selected.

screenshot from 2019-02-15 08-58-17

PhilipCattanach:
Tested successfully after running an analysis using the 40 Sample Apps

reporting/impl/src/main/resources/reports/templates/techReport-punchCard.ftl
```diff
@@ -292,9 +305,70 @@
...
+                $(tdElement).removeClass("sort_desc");
+            });
+
+            if (sector == currentSortSector) {
+                reverseSector = !reverseSector;
+            } else {
+                reverseSector = false;
+                currentSortSector = sector;
+            }
+            var f = reverseSector ? 1 : -1;
```
jonathanvila: in order to read faster, would it be nicer having proper names for vars ?

jonathanvila: Apart from this minor suggestion, looks ok for me.

m-brophy:
I'll put in some meaningful variable names before it gets merged

m-brophy:
variable names changed as per @jonathanvila  review

PR #1348, opened 2019-03-01
WINDUPRULE-378: parameterized rule|project|artifactId bugfix

m-brophy:
A rule with a `when` condition like
```xml
    <when>
        <project>
            <artifact groupId="org.hibernate" artifactId="{type}" />
        </project>
    </when>
```
was not working due to this bug.

### DIFF
--- a/rules-java-project/addon/src/main/java/org/jboss/windup/project/condition/Artifact.java
+++ b/rules-java-project/addon/src/main/java/org/jboss/windup/project/condition/Artifact.java
@@ -98,7 +98,7 @@ public class Artifact implements Parameterized
     {
         Set<String> result = new HashSet<>();
         if (groupId != null) result.addAll(groupId.getRequiredParameterNames());
-        result.addAll(artifactId.getRequiredParameterNames());
+        if (artifactId != null) result.addAll(artifactId.getRequiredParameterNames());
         return result;
     }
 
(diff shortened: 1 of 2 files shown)


rules-java-project/addon/src/main/java/org/jboss/windup/project/condition/Artifact.java
```diff
@@ -96,7 +96,9 @@ public Version getVersion()
     @Override
     public Set<String> getRequiredParameterNames()
     {
-        return new HashSet<>(Arrays.asList("groupId", "artifactId"));
+        Set<String> result = new HashSet<>(groupId.getRequiredParameterNames());
```
mrizzi:
```suggestion
        Set<String> result = new HashSet<>();
        if (groupId != null) result.addAll(groupId.getRequiredParameterNames());
```

rules-java-project/addon/src/main/java/org/jboss/windup/project/condition/Artifact.java
```diff
@@ -96,7 +96,9 @@ public Version getVersion()
...
+        result.addAll(artifactId.getRequiredParameterNames());
```
mrizzi:
```suggestion
        if (artifactId != null) result.addAll(artifactId.getRequiredParameterNames());
```

PR #1350, opened 2019-03-22
WINDUP-2337: tech report jax-rs moved from http to rest category

m-brophy:
NB: this pr works in conjunction with https://github.com/windup/windup-rulesets/pull/371

### DIFF
--- a/reporting/impl/src/main/resources/org/jboss/windup/reporting/rules/generation/techreport/techReport-hierarchy.tags.xml
+++ b/reporting/impl/src/main/resources/org/jboss/windup/reporting/rules/generation/techreport/techReport-hierarchy.tags.xml
@@ -88,7 +88,6 @@
             </tag>
             <tag name="techBox:http" title="HTTP" pseudo="true" parents="techRow:java-ee" icon="Elevator-UpDown">
                 <tag ref="jax-ws"/>
-                <tag ref="jax-rs"/>
             </tag>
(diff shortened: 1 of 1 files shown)


reporting/impl/src/main/resources/org/jboss/windup/reporting/rules/generation/techreport/techReport-hierarchy.tags.xml
```diff
@@ -99,8 +98,9 @@
             <tag name="techBox:webservice" title="WebService" pseudo="true" parents="techRow:embedded" icon="Global_Ecosystem">
                 <tag ref="webservices"/>
             </tag>
-            <tag name="techBox:rest" title="REST" pseudo="true" parents="techRow:embedded" icon="Science_Circuit">
+            <tag name="techBox:rest" title="REST" pseudo="true" parents="techRow:embedded, techRow:java-ee" icon="Science_Circuit">
```
jonathanvila: I see this is adding java-ee as parent for jax-rs, but also for rest that before it didnt have this parent.
Is this a wanted effect ?

m-brophy: Yes, we want in the case of Jersey for instance, for the count to go in 'Embedded/REST' but for Jax-RS, for it to go in JAVAEE/REST

PR #1354, opened 2019-04-03
WINDUP-2351: exportCSV new file with tags listed by application

m-brophy:
This handles the production of the new file which can be found in the report folder.

### DIFF
--- a/reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
+++ b/reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
@@ -19,6 +19,7 @@ import org.jboss.windup.graph.model.LinkModel;
 import org.jboss.windup.graph.model.ProjectModel;
 import org.jboss.windup.graph.model.WindupConfigurationModel;
 import org.jboss.windup.graph.model.resource.FileModel;
+import org.jboss.windup.graph.traversal.OnlyOnceTraversalStrategy;
 import org.jboss.windup.graph.traversal.ProjectModelTraversal;
 import org.jboss.windup.reporting.category.IssueCategoryModel;
 import org.jboss.windup.reporting.config.classification.Classification;
@@ -194,38 +195,50 @@ public class ExportCSVFileRuleProvider extends AbstractRuleProvider
 
             ApplicationReportService applicationReportService = new ApplicationReportService(event.getGraphContext());
             TechnologyTagService techTagService = new TechnologyTagService(event.getGraphContext());
-            final List<ApplicationReportModel> apps = applicationReportService.findAll();
-            Map<ApplicationReportModel,List<TechnologyTagModel>> appsAndTechTags;
-            List<TechnologyTagModel> masterTagList = techTagService.findAll().stream().sorted((o1,o2) ->
+            List<ApplicationReportModel> apps = applicationReportService.findAll();
+            List<TechnologyTagModel> masterTagList = techTagService.findAll().stream().filter(tag -> !tag.getName().equals("Decompiled Java File")).sorted((o1,o2) ->
 
                             {
                                 String n1 = o1.getName();
                                 String n2 = o2.getName();
-                                return n1.compareTo(n2);
+                                return n1.compareToIgnoreCase(n2);
                             }).collect(Collectors.toList());
 
             List<String> headerFieldsList = new ArrayList<>();
             headerFieldsList.add("App Name");
+            ArrayList<String> appNames = new ArrayList<>();
+            ArrayList<ApplicationReportModel> distinctApps = new ArrayList<>();
+
+            apps.forEach(app ->
+            {
+                if(!appNames.contains(app.getProjectModel().getRootFileModel().getFileName())) {
+                    appNames.add(app.getProjectModel().getRootFileModel().getFileName());
+                    distinctApps.add(app);
+                }
+            });
+
+
 
             masterTagList.forEach( masterTag -> headerFieldsList.add(masterTag==null?"":masterTag.getName()));
             String[] headerStringsToWrite = headerFieldsList.toArray(new String[0]);
 
-            apps.stream().forEach((ApplicationReportModel app) ->
+            distinctApps.stream().sorted((o1,o2) ->
             {
-                ProjectModelTraversal traversal = new ProjectModelTraversal(app.getProjectModel());
+                String n1 = o1.getProjectModel().getRootFileModel().getFileName();
+                String n2 = o2.getProjectModel().getRootFileModel().getFileName();
+                return n1.compareToIgnoreCase(n2);
+            }).forEachOrdered(app ->
+            {
+                ProjectModelTraversal traversal = new ProjectModelTraversal(app.getProjectModel(),new OnlyOnceTraversalStrategy());
                 ArrayList<TechnologyTagModel> tagsForApp = new ArrayList<>();
                 ArrayList<TechnologyTagModel> structuredTagsForApp = new ArrayList<>();
 
-                Iterable<TechnologyTagModel> tagsForAppIterable = ((TreeSet<TechnologyTagModel>)techTagService.findTechnologyTagsForProject(traversal)).stream().sorted((o1,o2) ->
-
+                ((TreeSet<TechnologyTagModel>)techTagService.findTechnologyTagsForProject(traversal)).stream().filter(tag -> !tag.getName().equals("Decompiled Java File")).sorted((o1,o2) ->
                 {
                     String n1 = o1.getName();
                     String n2 = o2.getName();
-                    return n1.compareTo(n2);
-                }).collect(Collectors.toList());
-
-
-                tagsForAppIterable.forEach(tagsForApp::add);
+                    return n1.compareToIgnoreCase(n2);
+                }).forEachOrdered(tagsForApp::add);
 
                 masterTagList.forEach( masterTag ->
                 {
@@ -237,14 +250,14 @@ public class ExportCSVFileRuleProvider extends AbstractRuleProvider
                     {
                         structuredTagsForApp.add(null);
                     }
-                    headerFieldsList.add(masterTag==null?"":masterTag.getName());
+                    headerFieldsList.add(masterTag==null?"":masterTag.getName() + (masterTag.getVersion()== null?"":" " + masterTag.getVersion()));
                 });
 
                 List<String> appNameAndTagNames = new ArrayList<>();
                 appNameAndTagNames.add(app.getProjectModel().getRootFileModel().getFileName());
                 structuredTagsForApp.forEach( tag ->
                 {
-                    appNameAndTagNames.add(tag==null?"":tag.getName());
+                    appNameAndTagNames.add(tag==null?"":tag.getName() + (tag.getVersion()== null?"":" " + tag.getVersion()));
                 });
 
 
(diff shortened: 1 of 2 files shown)


PhilipCattanach:
I have reviewed this PR from a functional perspective and there are some issues to be addressed detailed within the [Jira](https://issues.jboss.org/browse/WINDUP-2351?focusedCommentId=13717729&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13717729)

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
```diff
@@ -183,50 +188,140 @@ private String buildLinkString(Iterable<LinkModel> links)
...
+            {
+                if(!appNames.contains(app.getProjectModel().getRootFileModel().getFileName())) {
+                    appNames.add(app.getProjectModel().getRootFileModel().getFileName());
+                    distinctApps.add(app);
+                }
+            });
+
+
+
+            masterTagList.forEach( masterTag -> headerFieldsList.add(masterTag==null?"":masterTag.getName()));
```
jonathanvila: Following the code I don't see a way where masterTag could be a null

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
```diff
@@ -183,50 +188,140 @@ private String buildLinkString(Iterable<LinkModel> links)
...
+            String[] headerStringsToWrite = headerFieldsList.toArray(new String[0]);
+
+            distinctApps.stream().sorted((o1,o2) ->
+            {
+                String n1 = o1.getProjectModel().getRootFileModel().getFileName();
+                String n2 = o2.getProjectModel().getRootFileModel().getFileName();
+                return n1.compareToIgnoreCase(n2);
+            }).forEachOrdered(app ->
```
jonathanvila: if this *forEachOrdered* applied to a stream coming from *sorted*, redundant ?

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
```diff
@@ -183,50 +188,140 @@ private String buildLinkString(Iterable<LinkModel> links)
...
+
+                masterTagList.forEach( masterTag ->
+                {
+                    if (tagsForApp.contains(masterTag))
+                    {
+                        structuredTagsForApp.add(masterTag);
+                    }
+                    else
+                    {
+                        structuredTagsForApp.add(null);
```
jonathanvila: Assigning a NULL to a List ?

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
```diff
@@ -183,50 +188,140 @@ private String buildLinkString(Iterable<LinkModel> links)
...
+            if (isLineForAppTagFile)
+            {
+                projectToFile.get(APP_FILE_TECH_CSV_FILENAME).writeNext(line);
+            }
+            else
+            {
+                projectToFile.get(projectModel.getName()).writeNext(line);
+                //Convert line array to ArrayList, add extra field for merged file on the end,
+                // then convert back to array to send to CSVWriter
+                ArrayList<String> mergedList = new ArrayList<String>(Arrays.stream(line).collect(Collectors.toList()));
```
jonathanvila: is not Collectors.toList() returning an ArrayList<String> ?

jonathanvila:
I see that there are no tests involved in this change, and taking a look to the current ExportCsvTest it only checks there's a generated file but not its content.

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
m-brophy: I agree - null catch removed

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
m-brophy: The intention is that the sorted group is returned in sorted order. forEachOrdered guarantees this but forEach doesn't was my interpretation of these methods

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
m-brophy: the requirement is to create a header row with each tag used in the project listed in alphabetic order. Each application then has their tags listed to match these columns. When an application doesn't contain a particular tag null is added to the application's structured tag list at the index corresponding to that of the tag in the header row so that later tags appear in the correct columns

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
m-brophy: Removing the cast causes this error: 

> Incompatible types. Required ArrayList<String> but 'collect' was inferred to R: no instance(s) of type variable(s) T exist so that List<T> conforms to ArrayList<String>

>

jonathanvila: I see.
But I think that using the interface would be eough : 
List<String> mergedList = Arrays.stream(line).collect(Collectors.toList());
Or specifying to type directly on the stream:
ArrayList<String> mergedList = Arrays.stream(line).collect(toCollection(ArrayList::new));

But I know it's only a matter of taste.... doing the new ArrayList doesn't do any harm.

reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
jonathanvila: I thought that *sorted* and then *forEach* was already using the sort and that *forEachOrdered* was meant for parallel streams ....

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
```diff
@@ -23,11 +23,15 @@ public void exportCsv() throws IOException {
...
 
         String csvContent = new String(Files.readAllBytes(csv.toPath()), "UTF-8");
-
         assertTrue(csvContent.contains("Windup1x-javaee-example-tiny.war"));
+
+        File allIssuesCsv = new File(tmp.getRoot(), "AllIssues.csv");
+        assertTrue(allIssuesCsv.exists());
+
+        File appTagsCsv = new File(tmp.getRoot(), "ApplicationFileTechnologies.csv");
+        assertTrue(appTagsCsv.exists());
```
jonathanvila: Probably I haven't understood the problem....but, was not the goal of the ticket to provide a sort on the tags in the CSV file generated ? where are we checking that the content of the CSV file is following that ?

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
```diff
@@ -33,5 +39,40 @@ public void exportCsv() throws IOException {
 
         File appTagsCsv = new File(tmp.getRoot(), "ApplicationFileTechnologies.csv");
         assertTrue(appTagsCsv.exists());
+
+        List<List<String>> records = readCSVFile(appTagsCsv);
+        records.forEach(record ->
```
jonathanvila: it's ok but as a suggestion only : 
assertTrue(records.allMatch(this::isRowSorted))

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
```diff
@@ -33,5 +39,40 @@ public void exportCsv() throws IOException {
...
+        );
+    }
+
+    private List<List<String>> readCSVFile(File csvFile) throws IOException
+    {
+        List<List<String>> rows = new ArrayList<>();
+        try (BufferedReader buffer = new BufferedReader(new FileReader(csvFile))) {
+            String lineString;
+            while ((lineString = buffer.readLine()) != null) {
+                String[] columns = lineString.split(DELIMITER);
```
jonathanvila: I thought it was not that easy to parse a CSV .
What happens if we have : 
35, 1456, "mytext", 89000, "other,text", 45
I mean the text encloser " " should allow to have the DELIMITER inside the text ?

jonathanvila: Here there's one approach that could help :
https://www.mkyong.com/java/how-to-read-and-parse-csv-file-in-java/

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
```diff
@@ -33,5 +39,40 @@ public void exportCsv() throws IOException {
...
+        //This loop ignores the first entry in the list (ie where i=0) which is the app name,
+        //only the entries after that are sorted
+        for(int i=1;i<row.size()-1;i++)
+        {
+            if(row.get(i).compareTo(row.get(i+1))>0)
+            {
+                return false;
+            }
+        }
+        return true;
```
jonathanvila: Wdyt about using a java 8 method to check lists ?
`public boolean isRowSorted(List<String> info) {
    return info.equals(info.stream().sorted().collect(Collectors.toList()));
}`

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
m-brophy: I like this, suggested change has been  made

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
m-brophy: We can't use exactly this as it was written because the first item in the row is always the app name. After that they are sorted.  I've altered the readCsv method to remove the first item from each row so the sort check can work like this.

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
m-brophy: My only point is this is a test class, the csv we input to it is always the same one, and we know none of the fields are like this. If you think that remembering this it's still worth doing, then I will.

jonathanvila: Ohh, ok, if this is a controlled behavior, then it's OK.

tests/src/test/java/org/jboss/windup/tests/bootstrap/migrate/ExportCsvTest.java
jonathanvila: `public boolean isRowSorted(List<String> info) {
    return info.equals(info.stream().skip(1).sorted().collect(Collectors.toList()));
}`

jonathanvila: But it's ok with your approach.

PR #1355, opened 2019-04-09
WINDUP-2032 Sorting by aggregated columns within a tier

m-brophy:
https://issues.redhat.com/browse/WINDUP-2032

PR #1361, opened 2019-04-25
WINDUP-2372: bubble report sectorStats column sorting not working

m-brophy:
https://issues.redhat.com/browse/WINDUP-2372


carlosthe19916:
@m-brophy I've tested this PR locally and it seems to solve the problem I've reported here https://github.com/windup/windup/pull/1355#issuecomment-552395321 it works fine and without any problem.

PR #1362, opened 2019-05-09
Make depencencies minOccurs 0

wheelerlaw:
Self-explanatory. The default value for `minOccurs` is 1, and it doesn't seem like that is necessary here.


windupgithubbot1:

mrizzi:
retest

PR #1365, opened 2019-10-24
WINDUP-2460 Bump nexus-indexer-data version

mrizzi:
https://issues.jboss.org/browse/WINDUP-2460


PhilipCattanach:
Thanks Marco - this is the change we discussed earlier this week. The yy.mm.dd convention is good.

PR #1372, opened 2019-12-20
WINDUP-2525 <where> parameter in <artifact> condition doesn't work

mrizzi:
https://issues.redhat.com/browse/WINDUP-2525

PR with code to have ~~https://github.com/windup/windup-rulesets/pull/390~~ https://github.com/windup/windup-rulesets/pull/397 to work.

### DIFF
--- a/rules-java-project/tests/src/test/java/org/jboss/windup/project/operation/test/ProjectWithHintTest.java
+++ b/rules-java-project/tests/src/test/java/org/jboss/windup/project/operation/test/ProjectWithHintTest.java
@@ -33,9 +33,11 @@ import org.jboss.windup.graph.model.FileLocationModel;
 import org.jboss.windup.graph.model.ProjectDependencyModel;
 import org.jboss.windup.graph.model.ProjectModel;
 import org.jboss.windup.graph.model.resource.FileModel;
+import org.jboss.windup.graph.service.GraphService;
 import org.jboss.windup.project.condition.Artifact;
 import org.jboss.windup.project.condition.Project;
 import org.jboss.windup.reporting.config.Hint;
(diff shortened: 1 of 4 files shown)


rules-java-project/tests/src/test/java/org/jboss/windup/project/operation/test/ProjectWithHintTest.java
```diff
@@ -172,7 +193,25 @@ public void perform(GraphRewrite event, EvaluationContext context, FileLocationM
...
+                    .when(Project.dependsOnArtifact(Artifact.withArtifactId("{param}")))
+                    .perform(Hint.titled("Test {param}").withText("Test {param} Body").withEffort(42).and(addMatch))
+                    .where("param").matches("abc")
+                    .addRule()
+                    .when(Project.dependsOnArtifact(Artifact.withGroupId("{param}")))
+                    .perform(Hint.titled("Test {param}").withText("Test {param} Body").withEffort(42).and(addMatch))
+                    .where("param").matches("xyz")
+                    // deliberately not matching rule to have a evaluationStrategy.modelSubmissionRejected() case
+                    .addRule()
+                    .when(Project.dependsOnArtifact(Artifact.withGroupId("{param}").andArtifactId("{param}")))
```
m-brophy: i don't think I understand this. Is it a check for an artifact with the same groupId and artifactId? Or does it substitute in order from a list of parameters?

mrizzi: Yep, it searches for a dependency with the same value for groupId and artifactId and, as you can imagine, it won't be found so this tests the `modelSubmissionRejected()` method.

PR #1379, opened 2020-04-16
WINDUPRULE-488 Adding check if the same Hint has not been already added

mrizzi:
https://issues.redhat.com/browse/WINDUPRULE-488


mrizzi:
The issue has been fixed in https://github.com/windup/windup-rulesets/pull/436

PR #1398, opened 2020-10-14
WINDUP-2771 Excluded optional dependencies from analysis

mrizzi:
https://issues.redhat.com/browse/WINDUP-2771


mrizzi:
The `windup-core-PR-dependents-check` and `windup-core-PR-dependents-check-jdk11` were supposed to fail in this PR.
The failures are related to `springboot-mobile-to-quarkus-00000-test` test and it reflects the expected change in the dependency retrieval behaviour from windup.
The below screenshot has the _PR vs master_ comparison

Screenshot from 2020-10-15 09-23-00

It can be checked (on the left) that the `META-INF/maven/org.springframework.boot/spring-boot-autoconfigure/pom.xml` is not triggering the rule anymore since it has the `<optional>true</optional>` tag for the searched dependency which does NOT mean the dependency is used in the analyzed application.

PhilipCattanach:
@mrizzi - The code looks fine. A very straight forward change, but with massive benefits. 
The testing of the changed against the camunda-bpm-spring-boot-starter-example-war-2.0.0.war confirmed that <optional> dependencies were ignored in the POM files, and the rules looking for those dependencies didn't fire.

PR #1400, opened 2020-11-11
WINDUP-2591 Moved to DependentPhase to avoid interation with '--skipReports' option

mrizzi:
https://issues.redhat.com/browse/WINDUP-2591

`--skipReports` option works in [WindupProcessorImpl](https://github.com/windup/windup/blob/master/exec/impl/src/main/java/org/jboss/windup/exec/WindupProcessorImpl.java#L351-L371) on filtering out all the report related phases from the analysis.
In this PR I detached the `ExportCSVFileRuleProvider` from `PostReportGenerationPhase` (skipped if `--skipReports` is set) and attached it to the `DependentPhase` but placing it "in the same place" it was within the ordered list of analysis' phases: after the `PostReportGenerationPhase` has started and before the `AttachApplicationReportsToIndexRuleProvider` (like it was before this change)

### DIFF
--- a/reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
+++ b/reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java
@@ -3,7 +3,13 @@ package org.jboss.windup.reporting.export;
 import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeSet;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


PhilipCattanach:
@mrizzi - Hi Marco when you skipReports then the AllIssues.csv and ApplicationName.csv files are generated but the ApplicationFileTechnologies.csv is not. I could live with that to be honest (and we could document it as a low priority Jira). I think if a user is exporting to csv and skipping the reports then they want to manage the migration issues from the CSV. The ApplicationFileTechnologies.csv is unimportant.

mrizzi:
> @mrizzi - Hi Marco when you skipReports then the AllIssues.csv and ApplicationName.csv files are generated but the ApplicationFileTechnologies.csv is not. I could live with that to be honest (and we could document it as a low priority Jira). I think if a user is exporting to csv and skipping the reports then they want to manage the migration issues from the CSV. The ApplicationFileTechnologies.csv is unimportant.

That's really a good catch @PhilipCattanach :+1: 
Basically the fix worked when `eap` was one of the targets because the Tattletale reports were generated and that let the [`produceApplicationListCSV`](https://github.com/windup/windup/blob/fbd45d28d9ac77e38e8f3aa56678fac331e79491/reporting/impl/src/main/java/org/jboss/windup/reporting/export/ExportCSVFileRuleProvider.java#L199) method work.
Commit 46616eb let that method to work just relying on the input paths and so the applications analyzed.

PhilipCattanach:
Retested successfully. All Good.

m-brophy (approved):

PR #1406, opened 2021-01-15
WINDUP-2949 Bump Maven index version to 21.01.14.Final

mrizzi:
https://issues.redhat.com/browse/WINDUP-2949


PhilipCattanach:
Tested with the Pet Clinic application.  Mandatory and Potential SP and incidents matched exactly. Slight discrepancy with spring-data-jpa and spring-data-commons not listed under "Embedded framework - Spring" in the Information category. But both of those artifacts did trigger rules under the Mandatory or Potential sections. So there is no concern about the index.

PR #1420, opened 2021-07-19
WINDUP-3114 Bump Maven Index version

mrizzi:
https://issues.redhat.com/browse/WINDUP-3114


PhilipCattanach:
Tested the CLI with the new index to ensure consistent outputs when executing the CLI with the same analysis config against the same test applications.

PR #1422, opened 2021-08-06
WINDUP-3142 Initial and basic GH Worflow for new build on JDK 11

mrizzi:
https://issues.redhat.com/browse/WINDUP-3142


mrizzi (commented):
Merging in order to compose it directly in next PRs to (try to) reach the build we want.

PR #1427, opened 2021-10-18
[WINDUP-3376] Diva integration

akihikot:
This PR adds rules-java-diva module for enabling diva transaction analysis.


windupgithubbot1:

mrizzi:
retest

rules-java-diva/addon/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.konveyor.tackle-diva</groupId>
+            <artifactId>diva-core</artifactId>
+            <version>0.1.1</version>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>com.ibm.wala</groupId>
```
mrizzi, line 31: It seems like many IBM Wala dependencies are used from [`diva-core`](https://github.com/konveyor/tackle-diva/blob/0.1.1/core/build.gradle#L38-L43): it would be better if `diva-core` could provided these IBM Wala dependencies to ensure the proper version is used.

rules-java-diva/addon/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+            <artifactId>com.ibm.wala.cast.java.ecj</artifactId>
+            <version>1.5.7</version>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.13.0</version>
+            <optional>true</optional>
```
mrizzi, line 65: This is another example of a dependency that should be provided from `diva-core` that needs it to run, without requiring this addon to "discover" which are the dependencies to be imported in order to make `diva-core` fully working.

mrizzi (commented):
Let's try to review this PR with a "divide&conquer" approach.
I'll try to review individually specific areas of the PR (pom.xml, tests, general Java coding style, etc) to allow us to work on different topics in parallel at the same time.
Hope it works and please tell me what do you think about this approach.

This is a first review of the pom.xml file which implies also some changes in `diva-core` dependencies management.

rules-java-diva/tests/src/test/java/org/jboss/windup/rules/apps/diva/DivaTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Test
+    public void testSpring() throws IOException {
+        try (GraphContext context = factory.create(true)) {
+            Path inputPath = Paths.get("src/test/resources/spring/");
+            Path outputPath = Paths.get(FileUtils.getTempDirectory().toString(), "Windup")
+                    .resolve(UUID.randomUUID().toString());
+
+            WindupConfiguration windupConfiguration = new WindupConfiguration().setGraphContext(context)
+                    .setOptionValue(SourceModeOption.NAME, true).addInputPath(inputPath).setOutputDirectory(outputPath);
```
mrizzi, line 56: I was expecting a new option (e.g. `enableDiva`) for enabling Diva but I can not see anything like this in tests.
Is it going to be added?

mrizzi (commented):
First tests review.

mrizzi (commented):
Some high level considerations on Java coding style worth evaluating against all the Java classes.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/service/GetOrCreateGraphService.java
```diff
@@ -0,0 +1,45 @@
...
+    public GetOrCreateGraphService(GraphContext context, Class<T> type) {
+        super(context, type);
+        try {
+            TYPE = (String) type.getField("TYPE").get(null);
+        } catch (IllegalArgumentException | IllegalAccessException | NoSuchFieldException | SecurityException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public T getOrCreate(String key, String value, String... kvs) {
```
mrizzi, line 24: It seems like this method is searching based on a set of key-value properties which is pretty close to what [`findAllByProperties`](https://github.com/windup/windup/blob/fca676af560e471fcd6631a2499bdf6cd7c596da/graph/api/src/main/java/org/jboss/windup/graph/service/GraphService.java#L83) method does.
I think we should try to leverage all the existing code base to reuse the code and, in case something is not implemented yet, contribute to the `GraphService` itself to improve just once and then let all the other components to leverage the improvements.
I feel like this class should be removed to leverage what `GraphService` or improving it.

mrizzi (commented):
Review to try to foster the usage of methods already available from interacting with the graph or improving them to accomplish new requirements.

rules-java-diva/addon/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+            <groupId>org.jooq</groupId>
+            <artifactId>joox</artifactId>
+            <version>1.2.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.konveyor.tackle-diva</groupId>
+            <artifactId>diva-core</artifactId>
+            <version>0.1.1</version>
+            <optional>true</optional>
```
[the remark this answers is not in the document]
akihikot: I thought diva-core and other dependencies represent 'impl' component of addons (which furnace loads by a separate class loader in my understanding).

rules-java-diva/addon/pom.xml
akihikot: I think windup components can omit versions either by windup-bom or parent module.  Do you think we can provide diva-bom from jitpack? (I don't think wala provides any bom...)

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/JanusGraphReport.java
```diff
@@ -0,0 +1,190 @@
...
+            if (model instanceof DivaOpModel && key.equals("rest-call")) {
+                DivaRestCallOpModel m = GraphService.addTypeToModel(gc, model, DivaRestCallOpModel.class);
+                this.model = (T) m;
+                builder.build(new Named<DivaRestCallOpModel>(gc, m));
+            }
+        }
+
+        @Override
+        public void put(String key, io.tackle.diva.Report.Builder builder) {
+            if (model instanceof DivaContextModel && key.equals("constraints")) {
```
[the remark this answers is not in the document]
akihikot: Thanks.  I will fix it, and do this also with other strings, e.g.,  "txid", "sql", for interfacing with diva's Report.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/JanusGraphReport.java
```diff
@@ -0,0 +1,190 @@
...
+                DivaStackTraceModel parent = null;
+                DivaStackTraceModel current = null;
+                for (Trace t : ((Trace) data).reversed()) {
+                    IMethod m = t.node().getMethod();
+                    SourcePosition p = null;
+                    try {
+                        p = m.getSourcePosition(t.site().getProgramCounter());
+
+                    } catch (InvalidClassFileException | NullPointerException e) {
+                    }
```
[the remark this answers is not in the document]
akihikot: I think these exceptions represent the case of a stacktrace element with unknown source location.  I will double-check if there are specific applications for which diva reports such stacktraces, and how we can handle this case properly inside windup.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/JanusGraphReport.java
```diff
@@ -0,0 +1,190 @@
...
+    public void add(T model) {
+        if (addEdge != null) {
+            addEdge.accept(model);
+        }
+    }
+
+    @Override
+    public void add(Builder builder) {
+        // TODO Auto-generated method stub
+
```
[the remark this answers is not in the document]
akihikot: The Report interface is an abstraction of builders for any Json object. Currently diva never puts strings and arrays inside arrays.  I'm leaving these methods for future extension.

rules-java-diva/addon/pom.xml
akihikot: The same answer as the last question.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/service/GetOrCreateGraphService.java
akihikot: Thanks.  I will check GraphService code and consider your suggestion.

rules-java-diva/addon/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.ibm.wala</groupId>
+            <artifactId>com.ibm.wala.util</artifactId>
+            <version>1.5.7</version>
```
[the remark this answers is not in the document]
akihikot: Okay I can handle this.

rules-java-diva/tests/src/test/java/org/jboss/windup/rules/apps/diva/DivaTest.java
akihikot: I will follow your suggestion, e.g., option's name, default value, etc.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/JanusGraphReport.java
```diff
@@ -0,0 +1,190 @@
...
+                m.setSql((String) value);
+
+            } else if (model instanceof DivaRestCallOpModel && key.equals("http-method")) {
+                ((DivaRestCallOpModel) model).setHttpMethod((String) value);
+
+            } else if (model instanceof DivaRestCallOpModel && key.equals("url-path")) {
+                ((DivaRestCallOpModel) model).setUrlPath(DivaLauncher.stripBraces((String) value));
+
+            } else if (model instanceof DivaRestCallOpModel && key.equals("client-class")) {
+
```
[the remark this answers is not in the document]
akihikot: I can fix the structure.

akihikot:
@mrizzi Thank your for the review.  It seems I can handle the code part of your suggestions.  Let me know if you have some more adivices about the pom.xml part, e.g., how to omit version numbers.

mrizzi:
> @mrizzi Thank your for the review. It seems I can handle the code part of your suggestions. Let me know if you have some more adivices about the pom.xml part, e.g., how to omit version numbers.

@akihikot I've opened 2 PRs to share some more advices on the `pom.xml` part:
1. https://github.com/konveyor/tackle-diva/pull/41
1. https://github.com/akihikot/windup/pull/2

akihikot:
@mrizzi Regarding `<optional>true</optional>' I think it's safer as I confirmed adding some code like in the test code would cause  
```
        ObjectWriter w = io.tackle.diva.Util.JSON_SERIALIZER;
        w.canSerialize(ObjectWriter.class);
```

```

**Comment by mrizzi** [comment]
retest

File: rules-java-diva/addon/pom.xml [hunk 11]
```diff
@@ -0,0 +1,112 @@
...
+
+    <name>Windup Rules - Java / DiVA - Basic</name>
+
+    <dependencies>
+        <!-- Local 3rd party Dependencies -->
+        <dependency>
+            <groupId>org.jooq</groupId>
+            <artifactId>joox</artifactId>
+            <version>1.2.0</version>
+        </dependency>
```
mrizzi, line 22: I found no reference to this library in the code and removing it allows the tests to run successfully anyway so I think we should remove it.
```suggestion
```

mrizzi (commented):
Further comment to remove a library that seems not needed.

rules-java-diva/tests/src/test/java/org/jboss/windup/rules/apps/diva/DivaTest.java
mrizzi: During the last meeting we decided together to use the `enableTransactionsReport` CLI option for enabling the Diva transactions report generation.

mrizzi:
retest

mrizzi:
> @mrizzi Thank you. Your changes worked properly. Regarding optional=true, I think it's safer to add it to reference to diva-core as I confirmed adding some code like
> 
> ```
>         ObjectWriter w = io.tackle.diva.Util.JSON_SERIALIZER;
>         w.canSerialize(ObjectWriter.class);
> ```
> 
> in the test code, would cause
> 
> ```
> java.lang.LinkageError: loader constraint violation: loader (instance of org/jboss/modules/ModuleClassLoader) previously initiated loading for a different type with name "com/fasterxml/jackson/databind/ObjectWriter"
> ```
> 
> while it doesn't compile with optional=true .

@akihikot everything should be fixed in https://github.com/akihikot/windup/pull/3: please give it a try testing it and check if it solves the issue for you.

akihikot:
@mrizzi Please check my changes that reflect your code-part reviews and requests.

mrizzi (commented):
retest

mrizzi:
test this please

mrizzi:
test this please

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/JanusGraphReport.java
```diff
@@ -0,0 +1,193 @@
...
+import com.ibm.wala.classLoader.IMethod;
+import com.ibm.wala.classLoader.IMethod.SourcePosition;
+import com.ibm.wala.shrikeCT.InvalidClassFileException;
+import com.ibm.wala.types.MethodReference;
+import com.ibm.wala.util.strings.StringStuff;
+
+import io.tackle.diva.Report;
+import io.tackle.diva.Trace;
+
+public class JanusGraphReport<T extends WindupVertexFrame> implements Report {
```
mrizzi, line 32: `JanusGraphReport` is too generic name and better change it to reflect its usage within the Diva context, so something like `DivaReport` might be better.
At the same time the suffix `Report` confuses me: isn't the Diva report supposed to be in https://github.com/windup/windup/pull/1434?

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/JanusGraphReport.java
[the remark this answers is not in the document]
mrizzi: any update on this?

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/service/DivaStackTraceService.java
```diff
@@ -0,0 +1,63 @@
...
+            traversal = traversal.not(__.out(DivaStackTraceModel.PARENT));
+        } else {
+            traversal = traversal.filter(__.out(DivaStackTraceModel.PARENT).is(parent.getElement()));
+        }
+//        long nanos = System.nanoTime();
+        DivaStackTraceModel model = getUnique(traversal);
+//        total += System.nanoTime() - nanos;
+//        if (++count % 100 == 0) {
+//            Util.LOGGER.info("count = " + count + ", time = " + (total / 1000000) + "ms");
+//        }
```
mrizzi, line 46: Please remove commented code.

mrizzi (commented):
As agreed, I made another round of review of code with further comments.
I'll continue to do some more investigations on the code because it takes time to get into the details but in the meantime I think it's worth sharing this first set of comments to let you move forward on working on this.

graph/api/src/main/java/org/jboss/windup/graph/service/GraphService.java
```diff
@@ -280,6 +280,34 @@ protected T getUnique(Traversal<?, ?> query)
         return frame((Vertex)resultObj);
     }
 
+    public T getOrCreate(String key, String value, String... kvs)
```
mrizzi, line 283: Since the method is based on properties and there's already a naming convention in place in this classe for `ByProperties` methods, I suggest to add such a prefix.
```suggestion
    public T getOrCreateByProperties(String key, String value, String... kvs)
```

mrizzi, line 283: Please add a test for this method.
`GraphServiceTest` seems a good place to add tests for this method.
For sure a test to check there's no duplication of model and just one is created.
Please also add a test when an odd number of parameters is provided in input because it looks like this code will break: it would be better to check `kvs` has even size as first thing and in case manage properly if it's odd.

mrizzi (commented):
Further comments on the code as I moved forward in the review.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,457 @@
...
+    public void perform(GraphRewrite event, EvaluationContext context) {
+
+        Boolean enableDiva = (Boolean) event.getGraphContext().getOptionMap()
+                .getOrDefault(EnableTransactionAnalysisOption.NAME, Boolean.FALSE);
+        if (!enableDiva) {
+            LOG.info("Skipping Diva analysis as " + EnableTransactionAnalysisOption.NAME + " option isn't set.");
+            return;
+        }
+        try {
+            Util.injectedCall(DivaIRGen.advices(), new String[] { "org.jboss.windup.rules.apps.diva.analysis" },
```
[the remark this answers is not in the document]
akihikot: This applies class-loading time code modification, specified in DivaIRGen.advices() by exploiting bytebuddy tool, to some WALA and eclipse JDT classes.  This is in order to customize IR generation for the purpose of Diva,  as one originally by WALA is too restrictive for our purpose.  I agree at some future our modifications needs to go directly into WALA project itself though.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,457 @@
...
+            clf = new ECJClassLoaderFactory(scope.getExclusions()) {
+                @Override
+                protected JavaSourceLoaderImpl makeSourceLoader(ClassLoaderReference classLoaderReference,
+                        IClassHierarchy cha, IClassLoader parent) {
+                    return new DivaSourceLoaderImpl(classLoaderReference, parent, cha, stdlibs);
+                }
+            };
+        }
+
+        DivaIRGen.init();
```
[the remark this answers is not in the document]
akihikot: IR here stands for Intermediate Representation. A quote from https://github.com/wala/WALA/blob/master/com.ibm.wala.core/src/main/java/com/ibm/wala/ssa/IR.java .

> The IR (Intermediate Representation) is the central data structure that represents the instructions of a particular method. The IR represents a method's instructions in a language close to JVM bytecode, but in an SSA-based register transfer language which eliminates the stack abstraction, relying instead on a set of symbolic registers. The IR organizes instructions in a control-flow graph of basic blocks, as typical in compiler textbooks.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaRuleProvider.java
```diff
@@ -0,0 +1,33 @@
...
+           .addRule()
+           .perform(new DivaLauncher());
+
+       String backupFile = System.getenv(BackupJanusGraph.JANUSGRAPH_BACKUP_DIR);
+       if (backupFile == null || backupFile.isEmpty()) {
+           return conf;
+       } else {
+           return conf.addRule()
+           .perform(new BackupJanusGraph());
+       }
```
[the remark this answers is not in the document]
akihikot: Please take a look at https://github.com/konveyor/tackle-diva/blob/main/Dockerfile.windup. We are using the above environment variable here.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/model/DivaEntryMethodModel.java
```diff
@@ -0,0 +1,11 @@
+package org.jboss.windup.rules.apps.diva.model;
+
+import org.jboss.windup.graph.model.TypeValue;
+import org.jboss.windup.rules.apps.java.model.JavaMethodModel;
+
+@TypeValue(DivaEntryMethodModel.TYPE)
+public interface DivaEntryMethodModel extends DivaConstraintModel, JavaMethodModel, DivaRestApiModel {
```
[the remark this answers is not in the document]
akihikot: Please take a look at https://github.com/konveyor/tackle-diva/tree/main/windup .  DivaRestApiModel is a common interface for both REST endpoint and REST client operation.  Currently diva recognizes quarkus's RestEasy clients https://quarkus.io/guides/rest-client for the latter.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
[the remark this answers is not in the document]
akihikot: Using this mechanism requires any class that accesses to WALA and JDT libraries by a designated class loader.  In the current code,  all such classes exist under org.jboss.windup.rules.apps.diva.analysis.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/service/DivaStackTraceService.java
```diff
@@ -0,0 +1,63 @@
...
+        if (parent == null) {
+            traversal = traversal.not(__.out(DivaStackTraceModel.PARENT));
+        } else {
+            traversal = traversal.filter(__.out(DivaStackTraceModel.PARENT).is(parent.getElement()));
+        }
+//        long nanos = System.nanoTime();
+        DivaStackTraceModel model = getUnique(traversal);
+//        total += System.nanoTime() - nanos;
+//        if (++count % 100 == 0) {
+//            Util.LOGGER.info("count = " + count + ", time = " + (total / 1000000) + "ms");
```
akihikot, line 45:
```suggestion
       DivaStackTraceModel model = getUnique(traversal);
```

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/JanusGraphReport.java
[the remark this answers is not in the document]
akihikot: We haven't yet identified specific examples that trigger the above exceptions.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,457 @@
...
+                                                if (a.getType().getName() == Constants.LJavaxWsRsPath) {
+                                                    model.setUrlPath(DivaLauncher.stripBraces(((ConstantElementValue) a
+                                                            .getNamedArguments().get("value")).val.toString()));
+                                                }
+                                                // @TODO. @WebServlet("/app")
+                                            }
+                                            cs.add(model);
+
+                                        } else if (c.category().equals(Report.HTTP_PARAM)) {
+                                            DivaRequestParamModel model = requestParamService.getOrCreate(
```
akihikot, line 285:
```suggestion
                                            DivaRequestParamModel model = requestParamService.getOrCreateByProperties(
```

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaRuleProvider.java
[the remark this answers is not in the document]
mrizzi: The `--keepWorkDirs` option might help you. The option is meant `to instruct MTA to not delete temporary working files, such as the graph database and extracted archive files.`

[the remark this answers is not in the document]
akihikot: Thanks your suggestion worked nicely.  https://github.com/konveyor/tackle-diva/pull/45

mrizzi (commented):
Based on some testing, it looks like the Diva addon identifies the transactions with the line of code where the prepared statement is created rather than where it gets executed.
For example in this block of a transaction
Screenshot from 2022-02-08 18-53-03
it looks like `TradeDirect (816, 33)` is where the execution happens but checking the application's code, in that line [TradeDirect.java#L816](https://github.com/WASdev/sample.daytrader7/blob/61437f63df2d61c2169570b324b98fbdd5ead2c9/daytrader-ee7-ejb/src/main/java/com/ibm/websphere/samples/daytrader/direct/TradeDirect.java#L816) there's the `PreparedStatement stmt = getStatement(conn, getQuoteForUpdateSQL);` code while the statement execution happens some lines below (at L819) where there's the `ResultSet rs = stmt.executeQuery();` line.

I think this should be properly reflected and the line number should be `819` and not `816`.

mrizzi (commented):
Further topic we discussed worth not forgetting about:

- A great enhancement would be to have Diva to manage also JAX-RS endpoints and, for example, the Java EE/ Jakarta EE `Transactional` annotation
- Transactions are hypothetical: static analysis doesn't take into account runtime conditional statements values. This could affect the queries executed in a transaction. In my test a line like `if (false) quoteData = getQuote(conn, symbol);` has been taken into account as part of the transaction even if it will never be executed. I think we should clarify in the report that transactions report are based on static analysis and might be not a perfect representation of how the application will work at runtime
- execution time: performance improvements have been already done but I have the impression while testing that the execution time increased a lot running the transaction report. I think we must keep also under control the performances to ensure the tool is optimized

rules-java-diva/addon/pom.xml
```diff
@@ -0,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.jboss.windup.rules.apps</groupId>
+        <artifactId>windup-rules-java-diva-parent</artifactId>
+        <version>5.2.2-SNAPSHOT</version>
```
mrizzi, line 10:
```suggestion
        <version>5.3.0-SNAPSHOT</version>
```

rules-java-diva/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.jboss.windup</groupId>
+        <artifactId>windup-parent</artifactId>
+        <version>5.2.2-SNAPSHOT</version>
```
mrizzi, line 11:
```suggestion
        <version>5.3.0-SNAPSHOT</version>
```

rules-java-diva/tests/pom.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>org.jboss.windup.rules.apps</groupId>
+        <artifactId>windup-rules-java-diva-parent</artifactId>
+        <version>5.2.2-SNAPSHOT</version>
```
mrizzi, line 10:
```suggestion
        <version>5.3.0-SNAPSHOT</version>
```

mrizzi (commented):
@akihikot proposed changes to move to `5.3.0-SNAPSHOT` version

akihikot:
> Based on some testing, it looks like the Diva addon identifies the transactions with the line of code where the prepared statement is created rather than where it gets executed. For example in this block of a transaction Screenshot from 2022-02-08 18-53-03 it looks like `TradeDirect (816, 33)` is where the execution happens but checking the application's code, in that line [TradeDirect.java#L816](https://github.com/WASdev/sample.daytrader7/blob/61437f63df2d61c2169570b324b98fbdd5ead2c9/daytrader-ee7-ejb/src/main/java/com/ibm/websphere/samples/daytrader/direct/TradeDirect.java#L816) there's the `PreparedStatement stmt = getStatement(conn, getQuoteForUpdateSQL);` code while the statement execution happens some lines below (at L819) where there's the `ResultSet rs = stmt.executeQuery();` line.
> 
> I think this should be properly reflected and the line number should be `819` and not `816`.

This issue is fixed by https://github.com/konveyor/tackle-diva/pull/57 and included in diva-core:0.1.4.

akihikot:
Version 0.1.5 of diva supports ear files analysis.

akihikot:
Explanations on  the above changes. 
1. The relation between `DivaStackTraceModel` and `FileLocationModel` changed from inheritance to edge (55d7df9) .  This is in order to fix broken indent issue with highlighting for `HintInlineModel` in source.ftl, when multiple hints are on the same file location. 
2. As a result, we needed to move the performance fix using `@Indexed` on the `DivaStackTraceModel` (https://github.com/konveyor/tackle-diva/issues/25#issuecomment-937528726) directly to `FileLocationModel` (b4f838d). 
3. Additionally, we noticed complex join query is faster using traversal from a fixed node via `GraphTraversalSource.V()` , rather than using `getRawTraversal()` (dffae06).

mrizzi (commented):
I report here some issues reported previously while waiting for having all the code for Diva merged in this single PR.

I created another test for being more productive in testing this and I've used the [`hibernate-tutorial-web-3.3.2.GA.war`](https://github.com/windup/windup/blob/master/test-files/hibernate-tutorial-web-3.3.2.GA.war) test application we already have in Windup .
As you can imagine, it's a (very old) sample application that has servlet and transactions, obviously using hibernate as the name suggests.
Running the analysis of that application with the transaction report enabled generate this exception during DivaLauncher execution:
```java
java.lang.NullPointerException
	at io.tackle.diva.Framework.getFilteredTargets(Framework.java:655)
	at io.tackle.diva.Framework.traverse(Framework.java:597)
	at io.tackle.diva.Framework.traverse(Framework.java:538)
	at org.jboss.windup.rules.apps.diva.analysis.DivaLauncher.launch(DivaLauncher.java:326
```
Then I tried also with the source code for that application (available from https://repo1.maven.org/maven2/org/hibernate/hibernate-tutorial-web/3.3.2.GA/hibernate-tutorial-web-3.3.2.GA-sources.jar) and in this case no exception has been thrown but no report has been created.
The refenced class `io.tackle.diva.Framework` is coming from diva-core so it's out of the PR and it's provided by the tackle-diva project itself.
Beside this I have to say that, despite Windup has never been a champion in logging, Diva is adding a lot of log with some entries that seem to be not very useful.
For example the log `INFO: TODO` from https://github.com/konveyor/tackle-diva/blob/0.1.5/core/src/main/java/io/tackle/diva/analysis/ServletAnalysis.java#L286 seems like something that should be removed.
Or log like `INFO: isTreeStructured = false` (from [https://github.com/windup/windup/blob/f4b6c1ad73b4ea04e89dff4cba0650391dfe4fe7/rul[…]ava/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java](https://github.com/windup/windup/blob/f4b6c1ad73b4ea04e89dff4cba0650391dfe4fe7/rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java#L161)) might be meaningful but if you think this is something the Windup's user could leverage, I feel like the message should be a little bit more verbose (and maybe moved to DEBUG level)
Now I won't report every single log entry but I think I gave some reasonable examples of improvements that could be done.

Another kind of log that got my attention is this one:
```
INFO: 1. [Moderate] Invalid class file at path /home/mrizzi/git/forked/windup/rules-java-diva/tests/target/diva-temp10172376756653030224/jrt/module-info.class
2. [SEVERE] class com.ibm.wala.classLoader.BytecodeClass$ClassNotFoundWarning : Lcom/syncleus/ferma/typeresolvers/TypeResolver
3. [SEVERE] class com.ibm.wala.classLoader.BytecodeClass$ClassNotFoundWarning : Ljava/awt/event/ActionListener
```
because it's reported as `INFO` but it mentions it's `SEVERE` which is confusing for users: is it an info or something severe that prevents the analysis from working?

I can see that `DivaLauncher` creates a diva-temp temp directory within the Java tmp dir. Windup already creates some temporary directories within the Windup output folder and it would be better to have Windup writing files only in the output folder.

To summarize:

1. Hibernate test application exception report
2. logs
3. temporary directory

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,583 @@
...
+        IClassHierarchy filteredCha = new FilteredClassHierarchy(cha, appClasses::contains);
+        IClassHierarchy relevantCha = new FilteredClassHierarchy(cha, relevantClasses::contains);
+
+        List<IMethod> entries = new ArrayList<>();
+        entries.addAll(ServletAnalysis.getEntries(filteredCha));
+        entries.addAll(SpringBootAnalysis.getEntries(filteredCha));
+        entries.addAll(QuarkusAnalysis.getEntries(filteredCha));
+
+        if (entries.isEmpty()) {
+            LOG.info("Diva: Found no entry methods for anlaysis");
```
mrizzi, line 307:
```suggestion
            LOG.info("Diva: Found no entry methods for analysis");
```

mrizzi (commented):
Just a fix for a typo I noticed while testing.

akihikot:
https://issues.redhat.com/browse/WINDUP-3377

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,613 @@
...
+        Map<String, Properties> appProps = new LinkedHashMap<>();
+        FileService files = new FileService(gc);
+        for (ProjectModel p : projects) {
+            String targetPath = Paths.get(p.getRootFileModel().getFilePath())
+                    .resolve("src/main/resources/application.properties").toString();
+            PropertiesModel file = (PropertiesModel) files.findByPath(targetPath);
+            if (file != null) {
+                try {
+                    appProps.put(p.getName(), file.getProperties());
+                } catch (IOException e) {
```
m-brophy, line 490: Perhaps log an error here so it doesn't just get swallowed and ignored?

m-brophy, line 490: Perhaps log an error here so it doesn't just get swallowed and ignored?

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,613 @@
...
+                                .traverse(
+                                        g -> g.has(FileModel.FILE_PATH, thePath).in(ProjectModel.PROJECT_MODEL_TO_FILE))
+                                .toList(ProjectModel.class);
+                        for (ProjectModel p : ps) {
+                            DivaAppModel app = toApp.apply(p);
+                            app.setEndpointName(e.getKey());
+                        }
+                    }
+                }
+            } catch (IOException e1) {
```
m-brophy, line 541: empty catch block

m-brophy, line 541: empty catch block

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/reporting/StackTraceToList.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public Object exec(@SuppressWarnings("rawtypes") List arguments) throws TemplateModelException {
+        if (arguments.size() != 1) {
+            throw new TemplateModelException("Error, method expects one argument");
+        }
+        if (arguments.get(0) instanceof DivaStackTraceModel) {
+            throw new TemplateModelException(
+                    "Error, method expects argument of type " + DivaStackTraceModel.class.getSimpleName());
```
m-brophy, line 34: I don't get this - the argument is an instance of DivaStackTraceModel but the error is that the method expects an argument of DivaStackTraceModel. I assume I'm missing something....

m-brophy, line 34: I don't get this - the argument is an instance of DivaStackTraceModel but the error is that the method expects an argument of DivaStackTraceModel. I assume I'm missing something....

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/service/DivaStackTraceService.java
```diff
@@ -0,0 +1,109 @@
...
+public class DivaStackTraceService extends GraphService<DivaStackTraceModel> {
+
+    FileLocationService fileLocationService;
+
+    public DivaStackTraceService(GraphContext context) {
+        super(context, DivaStackTraceModel.class);
+        fileLocationService = new FileLocationService(context);
+    }
+
+    // public static int count0 = 0;
```
m-brophy, line 30: remove commented out code from this class

m-brophy, line 30: remove commented out code from this class

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/reporting/StackTraceToList.java
akihikot: This is mistake.  We should check if `((StringModel)arguments.get(0)).getWrappedObject()`  is instance of `DivaStackTraceModel`.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,613 @@
...
+                    } else if (rootFileModel instanceof JarArchiveModel
+                            && Util.any(projects, p2 -> p2.getRootFileModel().asFile().getAbsolutePath()
+                                    .startsWith(unzippedPath.normalize().toString()))) {
+
+                        //scope.addToScope(JavaSourceAnalysisScope.SOURCE,
+                        //        new SourceDirectoryTreeModule(unzippedPath.toFile()));
+                        scope.addToScope(ClassLoaderReference.Application,
+                                new BinaryDirectoryTreeModule(unzippedPath.toFile()));
+
+                    } else if (rootFileModel instanceof JarArchiveModel) {
```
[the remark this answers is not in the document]
akihikot: I assume you mean the condition here. 
https://github.com/windup/windup/blob/7d3388110b3df4674005479fa1faf1fd1efb5245/rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java#L265-L267
This rule handles nested `.jar` files, e.g, spring-boot application jars.  Let us update the rule, as we noticed a corner case where the rule isn't triggered correctly (when all nested jar files are ignored).

decompiler/api/src/main/java/org/jboss/windup/decompiler/api/DecompilationListener.java
```diff
@@ -12,6 +12,13 @@
      */
     void fileDecompiled(List<String> sourceClassPaths, String outputPath);
 
+    /**
+     * Indicates that the files at inputPath has been decompiled to outputPath
+     */
+    default void fileDecompiled(List<String> sourceClassPaths, String outputPath, int[] lineMapping) {
```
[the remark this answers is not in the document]
akihikot: The parameter is used if someone overrides this default method, e.g., `rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/decompiler/FernflowerDecompilerOperation.java`.  Please also check the original PR that summarizes the relevant changes https://github.com/akihikot/windup/pull/10.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
```diff
@@ -0,0 +1,613 @@
...
+import org.jboss.windup.rules.apps.diva.model.DivaConstraintModel;
+import org.jboss.windup.rules.apps.diva.model.DivaContextModel;
+import org.jboss.windup.rules.apps.diva.model.DivaEndpointModel;
+import org.jboss.windup.rules.apps.diva.model.DivaEntryMethodModel;
+import org.jboss.windup.rules.apps.diva.model.DivaRequestConstraintModel;
+import org.jboss.windup.rules.apps.diva.model.DivaRequestParamModel;
+import org.jboss.windup.rules.apps.diva.model.DivaRestApiModel;
+import org.jboss.windup.rules.apps.diva.model.DivaRestCallOpModel;
+import org.jboss.windup.rules.apps.diva.service.DivaEntryMethodService;
+import org.jboss.windup.rules.apps.java.archives.model.IgnoredArchiveModel;
```
m-brophy, line 50:
```suggestion
import org.jboss.windup.graph.model.IgnoredArchiveModel;
```
Another PR that's just been merged has changed the location of this class

graph/tests/src/test/java/org/jboss/windup/graph/typedgraph/graphservice/GraphServiceTest.java
```diff
@@ -259,4 +259,38 @@ private void checkObject(TestFooModel created)
...
+                Assert.assertNull(foos[4].getProp1());
+                Assert.assertEquals("2", foos[4].getProp2());
+            } finally {
+                for (TestFooModel foo: foos) {
+                    if (foo != null)
+                        foo.remove();
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
akihikot: See 028f7f4.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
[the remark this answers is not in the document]
akihikot: See a8caed6.

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/service/DivaStackTraceService.java
akihikot: See 054568a

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/reporting/StackTraceToList.java
akihikot: See 054568a

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
akihikot: See c1ee067

rules-java-diva/addon/src/main/java/org/jboss/windup/rules/apps/diva/analysis/DivaLauncher.java
akihikot: See c1ee067

PR #1437, opened 2021-12-22
[WINDUP-3234] Avoid scanning maven "target" folders in sourceMode analysis

jmle:
See [jira ticket](https://issues.redhat.com/browse/WINDUP-3234).


carlosthe19916 (commented):
@jmle Thanks for this PR. I've added minor comments, could you please take a look at them? On the other hand, I've tested this PR and it works fine: the `target` folder is no longer analyzed.

This PR will work as long as the build directory is named `target`. Maven offers the posibility to change the `target` folder name using the following Maven settings:

```xml
<!-- Change the "target" folder name by "myDirectoryName" -->
<build>
    <directory>myDirectoryName</directory>
</build>
```
I know it is not common to do such configuration and having the `target` folder hardcoded should be enough for the majority of cases. Here are 2 things I consider we can discuss in the future:

- As you probably noticed Windup is "Maven centric"; There is no guarantee that Gradle projects will work as expected. For instance, this line of code is Maven Centric https://github.com/windup/windup/blob/9ae5d5dd67f46ada2980e190eb107e49d142f1ab/rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/service/WindupJavaConfigurationService.java#L186 . This is out of the scope of this PR I just wanted to mention it. So there is nothing to do in this PR regarding this topic.
- Before reviewing this PR I really thought that the [Excluding files](https://access.redhat.com/documentation/en-us/migration_toolkit_for_applications/5.1/html-single/cli_guide/index#exclude-files_cli-guide) capabilities of Windup using `.windup-ignore.txt` files were similar to `.gitignore` or `.dockerignore` files, where we can just use the ignore file as a vehicle to exclude files and folders. For instance, we use `.gitignore` for ignoring the target folder here https://github.com/windup/windup/blob/master/.gitignore#L8 . I've tested the [Excluding files](https://access.redhat.com/documentation/en-us/migration_toolkit_for_applications/5.1/html-single/cli_guide/index#exclude-files_cli-guide) feature of Windup and it doesn't work as `.gitignore` or `.dockerignore`. In my personal opinion, having a strong and similar way to `.gitignore` or `.dockerignore` is a more reliable way for allowing users to exclude files/folders from the analysis because we don't use hardcoded `target` folder names neither we are Maven Centric. I know it might be out of the scope of this PR so I'll leave it as an open question to discuss.

jmle:
@carlosthe19916 thanks for all the insight!

- I didn't know about the ability of changing the `target` folder name. Although it's an edge case, we might need to analyse it, and it's probably worth mentioning in our next standup.
- Regarding the ability to configure which files to ignore, in the case of a `target` directory for instance it wouldn't be as simple as just saying "ignore all target directories". That would make this hypothetical "ignore engine" more complex, since it would have to check for the existence of a pom file as a sibling of `target`. Not sure if that's something that end users would need, but it might also be worth discussing.

carlosthe19916:
> * Regarding the ability to configure which files to ignore, in the case of a `target` directory for instance it wouldn't be as simple as just saying "ignore all target directories". That would make this hypothetical "ignore engine" more complex, since it would have to check for the existence of a pom file as a sibling of `target`. Not sure if that's something that end users would need, but it might also be worth discussing.

Thanks for your reply @jmle . Perhaps I didn't explain myself, I'm not suggesting a "Ignore all target directories" approach. What I'm suggesting is a REGEX approach to define which files/folders to exclude. Let me write an example, let's say we have this folder structure:

```
├── pom.xml
├── src
│   └── main
│       ├── java
│       │   └── target
│       │       └── MyClass.java
│       └── resources
│           ├── application.properties
│           └── target
│               └── anyfile.xml
└── target
    ├── classes
    │   ├── application.properties
    │   ├── META-INF
    │   │   └── panache-archive.marker
    │   └── target
    │       ├── anyfile.xml
    │       └── MyClass.class
    ├── generated-sources
    │   └── annotations
    └── maven-status
        └── maven-compiler-plugin
            └── compile
                └── default-compile
                    ├── createdFiles.lst
                    └── inputFiles.lst
```

From the example above, the `target` folder is present 3 times:
- Is a java package
- Is inside the `resources` folder
- Is the build directory where the compiled code is written

Forget about Windup and imagine we want to write a `.gitignore` file to ignore only the `target` folder where the compiled files are located then the content of `.gitignore` would be:

```
/target
```

this way we only exclude the root `target` folder. If the content of `.gitignore` were `target/` then ALL `target` folders will be ignored (which is not what we want in this case). Notice the importance of the location of the character `/` (at the end or beginning) which defines the behavior of the REGEX.

My point is that there is always a way to write a  `.gitignore` file to exclude files and folders and this approach is wildly adopted and well known; it might be an option to use a similar approach to create the content of our `.windup-ignore.txt` files.

What do you think? Perhaps there is something I'm missing.

jmle:
@carlosthe19916 I see, of course in that case it would make things simpler for it to be implemented. I think it's definitely something we could discuss, but I still don't know if there's a need from the userbase to exclude certain files and folders from the analysis. Also bear in mind that, in the particular case of this PR, it's probably not something we would want to leave for the user to configure, since duplicate results don't look good at all...

In any case, if we ever wanted to let the user ignore certain files, your "windupignore" approach sounds like the way to go to me.

jmle:
@carlosthe19916 we have created [WINDUP-3243](https://issues.redhat.com/browse/WINDUP-3243) to track a new `.gitignore` approach which would add an option to ignore files present in the `.gitignore` file.

carlosthe19916:
retest

PhilipCattanach:
Hi @jmle I've tested the changes by analyzing an application with a specific custom rule that uses the project condition identify a particular dependency with the pom.xml. The rule only fired once regardless of whether there was, or wasn't, a target directory with a pom.xml nested within it.  I agree that Carlos's suggested enhancements can be managed separately and what you have done covers the scope of the Jira issue. 
Thank you. 
I'll let @carlosthe19916  merge the change when he has happy and the automated checks have completed.

carlosthe19916:
Thanks for the PR @jmle now that the CI tests passed I will merge this PR. This PR does what was requested in the Ticket and following enhancements can be done using the Ticket created at https://issues.redhat.com/browse/WINDUP-3243

PR #1439, opened 2021-12-23
[WINDUP-3245] General refactoring

jmle:
Refactoring of some main classes improving readability and organisation:
- `TechnologyReferenceTransformer*` -> `TechnologyReferenceAliasTranslator*`. Could be further renamed to `TechnologyAliasTranslator`. I think it better tells what this set of classes do. Further refactoring would potentially be needed to also change the name of the xml files that contain the translation.
- `RuleLoaderImpl`, `WindupProcessorImpl`: extraction of blocks of code into methods to better show what they do.
- Rewriting of some expressions into lambdas.
- Renaming of some variables.
- Elimination of dead code.


m-brophy (commented):
I tested this by running against a sample app then running exactly the same analysis with a version built from master and everything looked the same apart from the sorting of issues within a category with the same number of story points. I don't think that's an issue ( I'm not even sure if there is a sort at all, it's not alphabetical anyway)

m-brophy:
retest

m-brophy (approved):

PR #1444, opened 2022-01-12
[WINDUP-3213] Upgrade dom4j dependency

jmle:
See [WINDUP-3213](https://issues.redhat.com/browse/WINDUP-3213).


mrizzi (approved):
thanks for the enhancement @jmle :+1:

PR #1447, opened 2022-01-18
[WINDUP-3215] Specify chrome driver location

jmle:
See [this list of default variables present in workflow images](https://github.com/actions/virtual-environments/blob/ffa2110464ecda1cc62b477f29609525de24e570/images/linux/Ubuntu1804-README.md).

### DIFF
--- a/tests/src/test/java/org/jboss/windup/tests/application/WindupArchitectureMediumBinaryModeProcyonTest.java
+++ b/tests/src/test/java/org/jboss/windup/tests/application/WindupArchitectureMediumBinaryModeProcyonTest.java
@@ -4,6 +4,7 @@ import java.util.Properties;
 
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.windup.graph.GraphContext;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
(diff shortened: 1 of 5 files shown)


tests/src/test/java/org/jboss/windup/tests/application/WindupArchitectureMediumBinaryModeProcyonTest.java
```diff
@@ -1,26 +0,0 @@
-package org.jboss.windup.tests.application;
```
mrizzi, line 1: Are we removing the test because Procyon doesn't work anymore?

mrizzi (commented):
A doubt about that the removal of the `WindupArchitectureMediumBinaryModeProcyonTest` means.

tests/src/test/java/org/jboss/windup/tests/application/WindupArchitectureMediumBinaryModeProcyonTest.java
jmle: Sometime ago we had a conversation about this particular test and we decided to exclude it from execution. Since the test was already being excluded I decided to remove it. Not sure if procyon doesn't work anymore, although probably we can use Fernflower exclusively.

mrizzi (approved):
Thanks @jmle :+1: 

As agreed there's going to be another PR to fix the Ubuntu host's version to be the `latest` everywhere.

PR #1451, opened 2022-01-24
[WINDUP-3215] Set all jobs to use ubuntu-latest

jmle:
See [WINDUP-3215](https://issues.redhat.com/browse/WINDUP-3215)


mrizzi (approved):
LGTM, thank @jmle :+1:

PR #1459, opened 2022-02-24
[WINDUP-3284] Inconsistency between JDK8 and JDK11 when analyzing certain java classes

jmle:
See [WINDUP-3284](https://issues.redhat.com/browse/WINDUP-3284).


mrizzi (approved):
It has been a tough issue but I'm happy we worked together to get rid of this: thanks a lot @jmle, well done :+1:

PR #1463, opened 2022-03-07
[WINDUP-3236] Input apps should always be analyzed

jmle:


PhilipCattanach:
Thanks @jmle 

The code changes look fine.
 I have ran several tests without the AnalyzeKnownLibraries advanced option, comparing the results of a 521 analysis with a 530-SNAPSHOT and all of the differences are due to desired behaviour (not analyzing pom.xml files embedded within known libraries).

PR #1470, opened 2022-03-21
WINDUP-3307 Avoid 'project' value collision in the graph

mrizzi:
https://issues.redhat.com/browse/WINDUP-3307


PhilipCattanach:
@mrizzi - The fix worked fine when analyzing that brsear.ear application the only failure in the rules is java-glassfish-groovy-01000 and the application details report is now present.

PR #1474, opened 2022-03-31
Moved to 'FINE' and enhanced log about archive inspection

mrizzi:


jmle:
Looks good to me. Just to leave the info here - I think that we could move this block of code inside the `if` block further down so that the index is not queried more than once for archives that are already set to be analysed. Not for this release but probably good to consider for the next one.

PR #1475, opened 2022-04-05
Bump Lucene index version to 22.04.04.Final

mrizzi:


mrizzi (commented):
Builds worked so it's fine to be merged.
No changes in the index creation, just a standard index content update.

PR #1482, opened 2022-04-21
Be able to skip not readable entries in a Zip Stream

carlosthe19916:


utils/src/main/java/org/jboss/windup/util/ZipUtil.java
```diff
@@ -186,4 +186,18 @@ public static List<String> scanZipFile(String parentPath, InputStream is, boolea
         }
     }
 
+    /**
+     * Return the next Entry of a ZipInputStream. If the nextEntry can not be read then kip it and continue with the next one
+     * @param zis
+     * @return next zip entry
+     * @throws IOException
+     */
+    public static ZipEntry getNextEntry(ZipInputStream zis) throws IOException {
```
jmle, line 195: Does this method have to be public?

utils/src/main/java/org/jboss/windup/util/ZipUtil.java
```diff
@@ -186,4 +186,18 @@ public static List<String> scanZipFile(String parentPath, InputStream is, boolea
         }
     }
 
+    /**
+     * Return the next Entry of a ZipInputStream. If the nextEntry can not be read then kip it and continue with the next one
```
jmle, line 190: Typo in "kip"

jmle (changes requested):

utils/src/main/java/org/jboss/windup/util/ZipUtil.java
carlosthe19916: Changed to private

utils/src/main/java/org/jboss/windup/util/ZipUtil.java
carlosthe19916: Thanks. Fixed and now it says "skip"

jmle:
@carlosthe19916 there are conflicts in this branch - I'd recommend to fix them before it gets stale, and maybe we can merge it

jmle (approved):

PR #1484, opened 2022-05-10
[WINDUP-3366] Possibility to analyse dependencies in source mode

jmle:
See [https://issues.redhat.com/browse/WINDUP-3366](https://issues.redhat.com/browse/WINDUP-3366).


jmle:
@carlosthe19916 thank you for your insights! Regarding adding the "download dependencies" feature, I guess it's something that @rromannissen could be interested in, so maybe he can give some feedback about that. But regardless of that, I think that the title of the issue in Jira might be a bit confusing: from the point of view of windup, the only thing this does is to decompile and analyse Java class files found in a sourceMode analysis. The "source + dependencies" nomenclature is just something related to Tackle. Even though this feature has come as a request from Tackle, I think that it makes sense for the sourceMode to decompile and analyse class files, independently of how Tackle will use it, because I also think that source mode and binary mode should behave as similarly as possible to avoid confusion in for the users.

PR #1486, opened 2022-05-13
[WINDUP-3383] Detect dependencies also within <profiles> block

jmle:
This will detect dependencies also within the `<profiles>` block of a POM file. Bear in mind that no additional "location" values have been added to the "project" rule: an artifact with a location of `PLUGINS` can be detected inside and outside the profiles block.

See [Jira ticket](https://issues.redhat.com/browse/WINDUP-3383).


carlosthe19916:
Thanks @jmle this PR LGTM!

PR #1488, opened 2022-05-19
[WINDUP-3388] Fix Dependencies report when --analyzeKnownLibraries is false

jmle:
This PR modifies the behaviour of the POM processing. If an archive is known and has been set to be ignored, now it will still be added to the graph. This is so that its information is available at a later stage for the Dependencies report to be filled. Otherwise, the Dependencies report will be missing most of the information, which comes from the POMs. Also, moved `IgnoreArchiveModel` to be along `ArchiveModel`.

See [Jira ticket](https://issues.redhat.com/browse/WINDUP-3388).


rules-java/api/src/main/java/org/jboss/windup/rules/apps/java/scan/provider/DiscoverMavenProjectsRuleProvider.java
```diff
@@ -45,6 +45,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.StringJoiner;
+import java.util.logging.Level;
```
carlosthe19916, line 48: Please remove this line since this import is not needed

carlosthe19916 (commented):
Just a minor change. please see my comment below

PR #1490, opened 2022-05-24
WINDUP-3403: provide maven Central link for dependencies which do not…

m-brophy:
… contain their own pom file

https://issues.redhat.com/browse/WINDUP-3403


PhilipCattanach:
Thanks @m-brophy I have tested the change and it works great.
I'll hold off merging it with master until we've had a chance to show it to Ramon.
I can't see why he would object as it is an improvement.

PR #1492, opened 2022-05-30
[WINDUP-3364] - Generate java docs

carlosthe19916:
https://issues.redhat.com/browse/WINDUP-3364

To generate the docs you need to follow the instructions from https://github.com/windup/windup/wiki/Dev-Create-the-JavaDoc

Instead of executing the `validate` maven phase:
```
mvn validate -PjavadocDist -Djavadoc.branding="Windup $VER Javadoc"
```

Now you need to execute the `package phase`:
```
mvn package -PjavadocDist -DskipTests -Djavadoc.branding="Windup $VER Javadoc"
```

## Why the maven phase changed
- Old version of `maven-javadoc-plugin` v2 throws errors while generating the java docs
- Since v3 of `maven-javadoc-plugin` the goal `generate-jar` invokes the execution of the `compile` phase so we need to generate the docs after any `compile` phase. To se more details read https://maven.apache.org/plugins/maven-javadoc-plugin/aggregate-jar-mojo.html

> Note any `maven-javadoc-plugin` v3 will take longer time than v2 to generate the docs because it forces to invoke the `compile` command.


carlosthe19916:
> @carlosthe19916 looks good - do we need to change anything in our builds for this to work?

No, we don't need to change anything on our builds since so far the Javadocs are generated manually.

mrizzi (commented):
I've tried this PR on top of latest master and I got the following mistake:
```java
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.10:copy-dependencies (copy-dependencies) on project windup-bootstrap: Artifact has not been packaged yet. When used on reactor artifact, copy should be executed after packaging: see MDEP-187.
```
which points to https://issues.apache.org/jira/browse/MDEP-187.

Then, just for the sake of testing, I've tried commenting https://github.com/windup/windup/blob/2b96360d9f3d59bcb774fb13562d8025373583de/bootstrap/pom.xml#L167-L185 and then the `mvn package -PjavadocDist -DskipTests -Djavadoc.branding="Windup 0.0.0.Final Javadoc"` command ran but there's nothing in the expected `target/apidocs/` output folder.

carlosthe19916:
After latest changes in the PR, the way for generating the html java docs will be:

```
mvn install -DskipTests
mvn validate -PjavadocDist -DskipTests -Djavadoc.branding="Windup VERSION Javadoc"
```
 So the pages will be generated in 2 steps.
Temporarily I added a GH action `Javadocs`for verifying the commands work in GH and generates the HTML files. After one member of the team tests with PR I will remove that GH workflow

PR #1493, opened 2022-05-30
[WINDUP-3411]-Github Actions running on Windows and Macos

carlosthe19916:
https://issues.redhat.com/browse/WINDUP-3411


jmle:
@carlosthe19916 is there any reason why we're doing this only for the `dependants` workflows?

carlosthe19916:
Not needed any longer

PR #1494, opened 2022-05-31
[WINDUP-3404] Make binary mode accept source too, reorganise input handling

jmle:
This PR removes sourceMode restrictions when operating in "binary mode".

Dependent PR: [https://github.com/windup/windup-rulesets/pull/708](https://github.com/windup/windup-rulesets/pull/708)

See [Jira ticket](https://issues.redhat.com/browse/WINDUP-3404).


exec/api/src/main/java/org/jboss/windup/exec/configuration/WindupConfiguration.java
```diff
@@ -221,8 +230,7 @@ public WindupConfiguration addInputPath(Path inputPath)
      */
     public Collection<Path> getInputPaths()
     {
-        Collection<Path> inputPaths = getOptionValue(InputPathOption.NAME);
-        return inputPaths;
+        return getOptionValue(InputPathOption.NAME);
```
carlosthe19916, line 233: Also, no need to change anything in this file since the original code and the new one are equivalent. Do you want to get the `getOptionValue(InputPathOption.NAME);` result for debugging purposes?

exec/api/src/main/java/org/jboss/windup/exec/configuration/options/OutputPathOption.java
```diff
@@ -81,15 +86,14 @@ public static ValidationResult validateInputAndOutputPath(Path inputPath, Path o
...
     {
 
+        inputPaths.stream().forEach(
+                input -> LOG.log(Level.INFO, "[Jandepora] input.toString())")
+        );
+
         if (inputPaths == null)
         {
             return new ValidationResult(ValidationResult.Level.ERROR, "Input path must be specified.");
         }
```
carlosthe19916, line 96: Could you please double-check this block of code? There is a `if NULL` condition AFTER the variable is actually used. If the `inputPaths` is NULL then a NullPointException will appear in line 89 since it is used before it was checked.

carlosthe19916 (changes requested):
@jmle I've made a quick first round of review and I'm leaving some comments. No need to address those comments right now since I still need to test this PR together with the final CLI and check the expected behaviour

exec/api/src/main/java/org/jboss/windup/exec/configuration/WindupConfiguration.java
jmle: This and other changes (the import ones) are changes I do as I develop the solution in order to cleanup the code and leave it in a slightly better shape that I found.

bootstrap/src/main/java/org/jboss/windup/bootstrap/commands/windup/RunWindupCommand.java
```diff
@@ -42,6 +34,23 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
```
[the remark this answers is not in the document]
jmle: We try not to use `*` imports, as they can lead to confusion about what is being imported or not. This and other import changes address those concerns.

carlosthe19916 (approved):
Thanks for all the effort and time put in this PR @jmle . Multiple tests has been done by @PhilipCattanach and the time to merge this PR has come.

PR #1500, opened 2022-06-08
Skip Forge addon installation if skipTests is provided

mrizzi:
No need to run the "addon-install" goal if the tests are going to be skipped so better to skip it as well


mrizzi (commented):
Verified in [`windup-build`](https://github.com/windup/windup/runs/6796660170?check_suite_focus=true)  that `windup-tests` module executed:
```
2022-06-08T17:48:26.3907714Z Results :
2022-06-08T17:48:26.3907973Z 
2022-06-08T17:48:26.3909340Z Tests run: 135, Failures: 0, Errors: 0, Skipped: 4
```
as expected.

PR #1417, opened 2021-07-05
Fixed Eclipse deps' versions

mrizzi:
https://issues.redhat.com/browse/WINDUP-3112

PR #1502, opened 2022-06-15
Fixed some Eclipse dependencies version to avoid issues with latest releases

mrizzi:
The builds are failing from yesterday with the message:
```
[ERROR] Failed to execute goal on project windup-eclipse-jdt: Could not resolve dependencies for project org.jboss.windup.jdt:windup-eclipse-jdt:jar:5.3.1-SNAPSHOT: Failed to collect dependencies at org.eclipse.jdt:org.eclipse.jdt.core:jar:3.16.0 -> org.eclipse.platform:org.eclipse.core.resources:jar:3.13.200 -> org.eclipse.platform:org.eclipse.core.expressions:jar:3.5.100 -> org.eclipse.platform:org.eclipse.core.runtime:jar:3.12.0 -> org.eclipse.platform:org.eclipse.equinox.preferences:jar:3.10.0 -> org.osgi.service:org.osgi.service.prefs:jar:[1.1.0,1.2.0): No versions available for org.osgi.service:org.osgi.service.prefs:jar:[1.1.0,1.2.0) within specified range
```
The cause is the latest release (v 3.10.0) published yesterday for org.eclipse.platform:org.eclipse.equinox.preferences [1]: in the [pom.xml file](https://repo1.maven.org/maven2/org/eclipse/platform/org.eclipse.equinox.preferences/3.10.0/org.eclipse.equinox.preferences-3.10.0.pom) it declares this dependency:
```xml
<dependency>
  <groupId>org.osgi.service</groupId>
  <artifactId>org.osgi.service.prefs</artifactId>
  <version>[1.1.0,1.2.0)</version>
</dependency>
```
but this dependency doesn't exist in any public Maven repository (at the comment's date, 2022-06-15).
The "closest" Maven artifact to this one seems to be `org.osgi:org.osgi.service.prefs` (i.e. without `.service` in the groupId) which is available, with the expected versions [2].

The solution is all about avoiding `org.eclipse.platform:org.eclipse.equinox.preferences:3.10.0` artifact from being used, forcing the previous release (i.e. `3.9.100`) to be used instead.

[1] https://repo1.maven.org/maven2/org/eclipse/platform/org.eclipse.equinox.preferences/3.10.0/
[2] https://repo1.maven.org/maven2/org/osgi/org.osgi.service.prefs/


mrizzi:
Commit 6ff860c reverts the changes coming from #1417 because now we're building on JDK 11 so the cause that required those changes (Windup building on JDK 8) is no longer applicable.

PR #1509, opened 2022-08-10
Add MTR Theme

carlosthe19916:


reporting/impl/src/main/resources-mtr/reports/templates/include/about.ftl
```diff
@@ -0,0 +1,29 @@
+<dl class="dl-horizontal">
+    <#if reportModel.relatedResources.windupAbout.windupRuntimeVersion?has_content>
+        <dt>Generated By</dt>
+        <dd>${getWindupBrandName()} ${reportModel.relatedResources.windupAbout.windupRuntimeVersion}</dd>
+    </#if>
+
+    <dt>Website</dt>
+    <dd><a href="https://developers.redhat.com/products/mta/overview/">https://developers.redhat.com/products/mta/overview/</a></dd>
```
jmle, line 8: We should create a ticket pointing to these lines of code to change the links when there's an MTR website available, since MTR pointing to the "old" MTA is not strictly correct I think

pom.xml
```diff
@@ -637,6 +637,28 @@
...
+            <activation>
+                <property>
+                    <name>downstream</name>
+                    <value>mtr</value>
+                </property>
+            </activation>
+            <properties>
+                <distribution.brand.name>Migration Toolkit for Runtimes by Red Hat (MTR)</distribution.brand.name>
+                <distribution.brand.name-acronym>MTR</distribution.brand.name-acronym>
+                <distribution.brand.documentation.url>https://developers.redhat.com/products/mta/overview/</distribution.brand.documentation.url>
```
jmle, line 651: We should also add these references to MTA to the ticket mentioned in the other comment I've made.

PD: as discussed, we can probably open another ticket to use this in the TFL template, or we can use the same one mentioned above

jmle (approved):

reporting/impl/src/main/resources-mtr/reports/templates/include/about.ftl
carlosthe19916: Ticket created here https://issues.redhat.com/browse/WINDUP-3461 we can update the links once those are available

pom.xml
carlosthe19916: Ticket created https://issues.redhat.com/browse/WINDUP-3462

PR #1516, opened 2022-09-21
Divided 'tests' module testing from other modules testing

mrizzi:
divided `windup-build` into `windup-build-with-unit-tests-1`, `windup-build-with-unit-tests-2` and `windup-tests`.
`windup-tests` just tests the module the name reflects.
The other two job, `windup-build-with-unit-tests-X` are a way for dividing in a 1-hour each the jobs to cover the remaining tests.

In the end the waiting time for the tests will move from more than 3 hours to something about 1 hour.

PS: the failures in other builds are unrelated to this PR but caused by changes in https://github.com/windup/windup/pull/1515

### DIFF
--- a/.github/workflows/pr_build_jdk11.yml
+++ b/.github/workflows/pr_build_jdk11.yml
@@ -22,7 +22,7 @@ jobs:
       run: |
         echo "CHROMEWEBDRIVER=${CHROMEWEBDRIVER}" >> "$GITHUB_ENV"
     - name: Build on JDK 11
-      run: mvn clean install -DskipTests -B -U && mvn -B test -pl '!tests,!rules-java/tests,!rules-xml/tests,!config/tests' -Dwebdriver.chrome.driver=${{ env.CHROMEWEBDRIVER }}/chromedriver
+      run: mvn clean install -DskipTests -B -U && mvn -B clean install -pl '!tests,!rules-java/tests,!rules-xml/tests,!config/tests' -Dwebdriver.chrome.driver=${{ env.CHROMEWEBDRIVER }}/chromedriver
 
   windup-build-with-unit-tests-2:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_build_jdk11.yml
```diff
@@ -6,10 +6,43 @@ on:
...
+      with:
+        java-version: '11'
+        distribution: 'adopt'
+        java-package: jdk
+        cache: 'maven'
+    - name: Set up chromedriver variable
+      run: |
+        echo "CHROMEWEBDRIVER=${CHROMEWEBDRIVER}" >> "$GITHUB_ENV"
+    - name: Build on JDK 11
+      run: mvn clean install -DskipTests -B -U && mvn -B clean install -pl '!tests,!rules-java/tests,!rules-xml/tests,!config/tests' -Dwebdriver.chrome.driver=${{ env.CHROMEWEBDRIVER }}/chromedriver
```
jmle, line 25: Why do we have to run `install` twice? Wouldn't we save some time by running `test` the second time, or does that not work?

mrizzi: As discussed, I've just kept it since I was thinking it was related to some issue in creating this job.
Since it looks like there were no issue, let's switch to have `test` instead.

jmle:
Looks good :+1:

.github/workflows/pr_build_jdk11.yml
mrizzi: This triggered some failure, e.g. `java.io.FileNotFoundException: Source '/home/runner/work/windup/windup/forks/jdt/../../forks/gremlin-shaded/target/classes' does not exist` when running `org.jboss.windup.graph.duplicate.typevalue.DuplicateTypeValueTest`

PR #1517, opened 2022-09-22
[WINDUP-3410] Add source and target info to each hint

jmle:
See [JIRA ticket](https://issues.redhat.com/browse/WINDUP-3410).

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -438,6 +438,11 @@
                 <version>${version.lucene}</version>
             </dependency>
             <dependency>
+                <groupId>org.apache.lucene</groupId>
+                <artifactId>lucene-backward-codecs</artifactId>
+                <version>${version.lucene}</version>
+            </dependency>
+            <dependency>
                 <groupId>com.google.code.gson</groupId>
                 <artifactId>gson</artifactId>
                 <version>2.8.9</version>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -438,11 +438,6 @@
                 <version>${version.lucene}</version>
             </dependency>
             <dependency>
-                <groupId>org.apache.lucene</groupId>
-                <artifactId>lucene-backward-codecs</artifactId>
-                <version>${version.lucene}</version>
-            </dependency>
-            <dependency>
                 <groupId>com.google.code.gson</groupId>
                 <artifactId>gson</artifactId>
                 <version>2.8.9</version>
--- a/config/impl/src/main/java/org/jboss/windup/config/loader/RuleLoaderImpl.java
+++ b/config/impl/src/main/java/org/jboss/windup/config/loader/RuleLoaderImpl.java
@@ -50,7 +50,7 @@ public class RuleLoaderImpl implements RuleLoader {
     }
 
     private RuleProviderRegistry buildRegistry(RuleLoaderContext ruleLoaderContext) {
-        List<Rule> allRules = new ArrayList<>(2000); // estimate of how many rules we will likely see TODO: careful with this
+        List<Rule> allRules = new ArrayList<>(2000); // estimate of how many rules we will likely see
 
         List<RuleProvider> providers = loadProviders(ruleLoaderContext);
         RuleProviderRegistry registry = new RuleProviderRegistry();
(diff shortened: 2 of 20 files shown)


carlosthe19916:
@jmle I have run the following command:
```
bin/windup-cli --input ~/apps/AdministracionEfectivo.ear --input ~/apps/jee-example-app-1.0.0.ear --target eap7 --overwrite
```
I consider there are things that we can enhance:

1. Not all the issues have a Source or Target label. I believe that the fact that there is an issue means that at least one rule matched the Source or Target inputs; therefore, at least one label must appear per each Issue.

Screenshot from 2022-11-21 14-07-39

2. The Target filter doesn't work. Watch the video below, although there is a valid issue with the filtered target , after applying the filter none of the issues that do match the filter are shown.
3. The Clear button doesn't work. 

[Screencast from 2022-11-21 14-13-49.webm](https://user-images.githubusercontent.com/2582866/203064452-09fc9158-c383-4fba-9bd6-c7d6d63a75f9.webm)

carlosthe19916 (changes requested):
@jmle Thanks for the enhancements, it looks better. Here are some points where we can enhance even more the Issues report:
- The Source and Target Filter Dropdowns are placed in the order "First Target" and "Second Source". In my opinion we can switch the order so we see the Dropdowns "Source" and then "Target". Just it is natural to think in "Source" => "Target" in that order. 
- We can select multiple items of the same Dropdown but there is no way of knowing which elements were selected already. For instance , if I select 2 elements in the dropdown "Target", both selected elements are part of the filter however it impossible to know which is the current set of selected elements in the dropdown. There might be multiple solutions, in the "Application List" page we face this issue having labels below the filter so the user knows which are the selected filters.

[Screencast from 2022-11-30 10-14-49.webm](https://user-images.githubusercontent.com/2582866/204755767-4a8e0871-afce-409c-b40c-881b0fb313d1.webm)

- It seems that the logic for multiple selections of the same dropdown is a AND condition. For instance, if I select 2 elements, option1 and option2, of the dropdown "Target" then the logic applied seems to be "Filter all issues that contains option1 AND option2". I believe the correct filter must be an OR condition rather than AND; otherwise as soon as the user selects 2 elements of the same dropdown none issue will be visible. So my suggestion is: for selections of the same dropdown use OR condition, and for selections of different dropdowns use AND. Eg. (source1 OR source2 or source3) AND (target1 OR target2 OR target 3)

What do you think @jmle ?

carlosthe19916:
@jmle Thanks for the latest changes. Your last commit fixed the Loading of scripts order in Chrome, well done! After testing it in Firefox and Chrome I'm approving this PR

carlosthe19916 (approved):

carlosthe19916:
Thanks for the enhancements @jmle . Just minor things I believe is worth enhancing:

- The "X" button that closes any selected Source/Target has color "Blue" and it makes difficult to see it. Could we set the button "X" with color White just as the "Application list" report does?
- The link "Clear" that clears all selected filters might be changed by "Clear All Filters" to be consistent with the text of the "Application list" report
- If I select more than 6 targets then all selected targets occupy 2 rows for showing them, but there is still enough space to print them in the same row. See the image below. I believe the issue comes from this line https://github.com/windup/windup/pull/1517/files#diff-3e882d3c659746c5624e239ac3d53b5760e5547ee500f0d6ac7847d2a54b6073R228 

Screenshot from 2023-01-20 11-40-57

carlosthe19916 (changes requested):
Thanks for the enhancements @jmle . Just minor things I believe is worth enhancing:

- The "X" button that closes any selected Source/Target has color "Blue" and it makes difficult to see it. Could we set the button "X" with color White just as the "Application list" report does?
- The link "Clear" that clears all selected filters might be changed by "Clear All Filters" to be consistent with the text of the "Application list" report
- If I select more than 6 targets then all selected targets occupy 2 rows for showing them, but there is still enough space to print them in the same row. See the image below. I believe the issue comes from this line https://github.com/windup/windup/pull/1517/files#diff-3e882d3c659746c5624e239ac3d53b5760e5547ee500f0d6ac7847d2a54b6073R228 

Screenshot from 2023-01-20 11-40-57

carlosthe19916 (approved):
Thanks @jmle after testing this PR, it looks good to me!

PhilipCattanach:
Retested yesterday and it worked well. Carlos is satisfied all of his PR feedback has been addressed. Happy to squash and merge.

PR #1518, opened 2022-10-04
Windup 3370 export summary file

m-brophy:
https://issues.redhat.com/browse/WINDUP-3370


jmle (changes requested):
Looks good to me, although I have one suggestion and one question:
1. Was it part of the contract of the story that the file should be generated in the root folder of the report? Because if not, it'd be good to place it under `reports/data`
2. When analysing with several input apps, there's an additional "application" in the summary called `shared-libs` with empty data on it. Is that expected?

```  {
    "application:": "shared-libs",
    "incidentsByCategory": {},
    "mandatoryIncidentsByType": {}
  }```

PR #1520, opened 2022-10-04
Upgrade janusgraph

carlosthe19916:


carlosthe19916:
The upgrade of Janusgraph will require more than just changing the version of artifact, closing this PR so more investigation can be done

PR #1548, opened 2022-10-11
Removed useless 'wicket-core' artifact test dependency declaration

mrizzi:
The `wicket-core` test dependency declaration is useless so it can be safely removed.


carlosthe19916 (approved):
LGTM! Thanks @mrizzi

PR #1570, opened 2022-11-29
Bump Nexus index 22.11.28.Final

mrizzi:


mrizzi (commented):
Having the build to work proves the artifact is available and fine with tests.

PR #1574, opened 2022-12-12
Bump ocpsoft/rewrite 3.5.1.Final

mrizzi:


jmle:
Do we know what the `Alpha9-jboss` version had in particular? I think that version was made specifically for windup, so I'm wondering if we will miss something by switching to a "normal" version. I have tried to find the branch for the version in the ocpsoft repo, but couldn't find it.

mrizzi:
> Do we know what the `Alpha9-jboss` version had in particular? I think that version was made specifically for windup, so I'm wondering if we will miss something by switching to a "normal" version. I have tried to find the branch for the version in the ocpsoft repo, but couldn't find it.

It's not a branch, it's a tag https://github.com/ocpsoft/rewrite/tree/3.0.0.Alpha9-jboss and it's available in the "normal" stream of tags of the upstream project (ref. https://github.com/ocpsoft/rewrite/tags)

PR #1575, opened 2022-12-13
Excluded 'org.eclipse.core.runtime' artifact from 'diva-core' dependency

mrizzi:

PR #1576, opened 2022-12-13
Excluded 'org.eclipse.core.runtime' artifact from 'diva-core' dependency

mrizzi:
Backport #1575


mrizzi (commented):
`master` branch PR already approved and the successful tests made this PR fine as well.

PR #1579, opened 2023-01-03
[WINDUP-3561] - Update static reports to PF4 - mandatory reports

carlosthe19916:
https://issues.redhat.com/browse/WINDUP-3561

https://github.com/windup/windup-distribution/pull/101 is required at the time of generating the CLI

## Reported issues

- [x] In the Project/Applications page, when you click on the number of incidents and a row with a breakdown of incidents by category appears, only Mandatory, Optional, Potential and Information category incidents are listed. Cloud-mandatory incidents for instance are not listed, so the total and sum of incidents listed do not match if these incidents are present in an analysis. Is it a design choice to only show the 'main' categories?
> Fixed: now all available Category Issues must appear.

- [ ] When you click on a particular application in the application list to view the analysis details for that application only and got to the Issues tab then filter by target, I notice if there's for example an item showing a target java-ee: [6,), if you click on the box in the filter drop down for java-ee 6, the item is still visible. If you click on java-ee 7 it is no longer visible although my understanding is the meaning of that target is java-ee 6 or above . Is that right?
- [x] In the Rule execution providers report, it might be worth adding a search by ID, as typically the user will want to see if a particular rule was fired or not.
> The report already contains a column called "ID" and you can use the search input box to type the ID of the rule you are searching for.

- [x] The header in the Technologies report should be "Technologies" and not "Dependencies"
- [x] The size of the page in pagination resets itself each time a different filter is selected - maybe it should remain the same
- [x] Sources/targets filters show all sources and targets, regardless of whether they have been triggered for the particular application or not. It should show only the ones triggered, otherwise it might get confusing
- [x] The Application messages section doesn't have a title
- [x] There is no mention of duplicate archives in the Application details section (see figure 3.7 [here](https://access.redhat.com/documentation/en-us/migration_toolkit_for_runtimes/1.0/html/cli_guide/review-reports_cli-guide)).
- [x] In the application dependencies area, for each dependency, "Found at path" only shows one path, instead of all the paths where the jar has been found
- [x] In the Application/Transactions tab page following a diva analysis, when you drill down into the source code page and click on 'View Hint' there isn't any useful information displayed in the right hand panel:
> @m-brophy The reason is because the hint the DIVA report generates is empty. The old and the new reports show nothing because it has not been design in that way. If you compare the old reports with the new ones you will notice both show nothing.

- [x] The Tags found - Occurrence Found histograms are empty on the new reports at both at Application and Archive level


m-brophy:
In the Project/Applications page, when you click on the number of incidents and a row with a breakdown of incidents by category appears, only Mandatory, Optional, Potential and Information category incidents are listed. Cloud-mandatory incidents for instance are not listed, so the total and sum of incidents listed do not match if these incidents are present in an analysis. Is it a design choice to only show the 'main' categories?

m-brophy:
When you click on a particular application in the application list to view the analysis details for that application only and got to the `Issues` tab then filter by target, I notice if there's for example an item showing a target `java-ee: [6,)`, if you click on the box in the filter drop down for `java-ee 6`, the item is still visible. If you click on `java-ee 7` it is no longer visible although my understanding is the meaning of that target is `java-ee 6 or above` . Is that right?

jmle:
In the Rule execution providers report, it might be worth adding a search by ID, as typically the user will want to see if a particular rule was fired or not.

jmle:
The header in the Technologies report should be "Technologies" and not "Dependencies"

jmle:
The size of the page in pagination resets itself each time a different filter is selected - maybe it should remain the same

jmle:
Sources/targets filters show all sources and targets, regardless of whether they have been triggered for the particular application or not. It should show only the ones triggered, otherwise it might get confusing

jmle:
The Application messages section doesn't have a title

jmle:
There is no mention of duplicate archives in the Application details section (see figure 3.7 [here](https://access.redhat.com/documentation/en-us/migration_toolkit_for_runtimes/1.0/html/cli_guide/review-reports_cli-guide)).

jmle:
In the application dependencies area, for each dependency, "Found at path" only shows one path, instead of all the paths where the jar has been found

m-brophy:
In the Application/Transactions tab page following a diva analysis, when you drill down into the source code page and click on View Hint there isn't any useful information displayed:
image

mrizzi (commented):
Reporting some issues discussed offline:

1. Themed reports:
check each themed report is properly generated
2. Output path of the reports:
remove the `pf-reports` folder in order to have the legacy and new reports to provide the `index.html` file in the same location so that every user/integration will keep find the reports where they have been so far
3. Sorting in the issues page:
legacy reports show the issues statically ordered with mandatory issues on top and information ones at the bottom. This should be consistently reproduced in the new reports. 
It would also very good to have the sorting available for `Total incidents` and `Total storypoints` columns as well to let the user easily and quickly identify the "worst" issue in each of those columns.

jmle:
@carlosthe19916 we need to make sure that the new reports are generated when running tests on windup-rulesets too

mrizzi:
> @carlosthe19916 we need to make sure that the new reports are generated when running tests on windup-rulesets too

I've done it in https://github.com/windup/windup-rulesets/pull/900

mrizzi (commented):
The incidents' descriptions in the old reports
Screenshot from 2023-03-15 11-32-58

and their different values in the new reports
Screenshot from 2023-03-15 11-33-41

mrizzi (commented):
@carlosthe19916 the change in the PR https://github.com/windup/windup/pull/1611 must be applied also to new reports here.

PR #1595, opened 2023-02-23
Restored version 6.1.3-SNAPSHOT

mrizzi:


mrizzi (commented):
Project's unit tests related builds are fine.
Dependent `windup-rulesets` project is not building but it's expected since it needs https://github.com/windup/windup-rulesets/pull/872 to be merged to work but it's dependent so it's a  "chicken or egg problem".

I'm going to merge this PR since tests are fine.

PR #1598, opened 2023-03-01
Fixed Maven to version '3.8.7' for 'windup-tests' job

mrizzi:


mrizzi (commented):
The scope of this change is to have the `windup-tests` job to run again since it has been broken but the latest Ubuntu GitHub runner version (ref. https://github.com/actions/runner-images/issues/7216).

Since the job is running, I'm going to merge this.

PR #1616, opened 2023-03-17
Fix 'windup-tests' job

mrizzi:
The issue has been triggered by the [Ubuntu 22.04 (20230313) Image Update](https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20230313.1) that adopted Chrome 111 with which Selenium has the issue https://github.com/SeleniumHQ/selenium/issues/11750.

To "properly" solve it, [Using Java 11+ HTTP Client in Selenium 4.5.0 and beyond](https://www.selenium.dev/blog/2022/using-java11-httpclient/) guide has been applied but it spotted another issue: it caused `java.lang.ClassNotFoundException: java.net.http.HttpTimeoutException` due to [MODULES-392](https://issues.redhat.com/browse/MODULES-392).
The latter issues was fixed in JBoss Modules `1.9.2` but Furnace `2.29.1.Final` is using `jboss-modules:1.9.1.Final`.
So the `netty` driver has been adopted.

Furthermore, changes from [Headless is Going Away!](https://www.selenium.dev/blog/2023/headless-is-going-away/) to prevent issues in the future together with Selenium dependencies updates.

### DIFF
--- a/test-util/pom.xml
+++ b/test-util/pom.xml
@@ -18,7 +18,7 @@
         <dependency>
             <groupId>org.seleniumhq.selenium</groupId>
             <artifactId>htmlunit-driver</artifactId>
-            <version>3.58.0</version>
+            <version>4.8.0</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.google.guava</groupId>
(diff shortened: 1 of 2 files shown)


mrizzi (commented):
The changes are "just" meant to have 'windup-tests' job to run again and it worked fine.
As explained in the PR's description there are no choices/decisions and the fix was kind of mandatory so I'm going to merge since it's a blocker for the many other open PRs we have.

PR #1619, opened 2023-03-24
Bump Nexus index 23.03.23.Final

mrizzi:


mrizzi (commented):
The build worked which proves the new index version is publicly available.

PR #1620, opened 2023-03-24
Added backport GH job

mrizzi:


mrizzi (commented):
The PR just adds the same `backport` job that other Windup's projects are using so I'm going to merge this.

PR #1623, opened 2023-03-24
WINDUP-3612 non-zero RC when failing an analysis

mrizzi:
https://issues.redhat.com/browse/WINDUP-3612

Here the idea is the the new `--exitCodes` (or `-e`) option works only when also `--batchMode` is provided based on the assumption that only a non-interactive execution (i.e. batch mode) could benefit from leveraging the exit code values.

~Depends on:~

- https://github.com/windup/windup-distribution/pull/116


mrizzi:
commit 82f2f60 removes the need for https://github.com/windup/windup-distribution/pull/116 in order to still build the CLI.

mrizzi:
@jortel please share here the outcome of your tests in order to approve these changes, thanks

jmle (approved):

mrizzi (commented):
@jortel reported:
"Tested using -e and with common failure scenarios I have seen.  In these cases, the windup CLI exit code was (1) as expected."

PR #1637, opened 2023-04-14
WINDUP-3793 ZIP creation: add 'logs' folder to excluded prefixes list

mrizzi:
https://issues.redhat.com/browse/WINDUP-3793


PhilipCattanach:
I have ran an analysis with the CLI and with the Web Ui having the exportZipReport option enabled.
In both cases the zip file is consistent in content. No log information in either. 
I will merge as soon as the automated tests complete.
Thanks!

PR #1651, opened 2023-04-27
WINDUP-3843 Fix source filter for issues table component

abrugaro:
https://issues.redhat.com/browse/WINDUP-3843
Fixes the source filter being ignored bug in the Issues Table component when a target filter is also provided.


mrizzi (approved):
Tested successfully
image

@abrugaro thanks a lot and congratulations on your first contribution :clap:

PR #1659, opened 2023-05-04
Hint: removed stacktrace printing for some warnings

mrizzi:
It can happen that some (supposed to be) parameters in hint's title and body have no values and this situation has been fully managed so far printing a warning followed by the exception's stacktrace.
It happened that some rules, perfectly working, managed to trigger this situation a lot of stacktraces were printed on the log for basically no reason considering the warning message already clarifies what happened.

So I've remove the stacktrace logging to avoid misunderstanding about this not being managed.

Then, since the same hint will generate the same log multiple times, I've managed to print it just once per hint (i.e. per rule) with the same approach to logging adopted also in other situation in Windup.


carlosthe19916 (approved):
LGTM! Thanks @mrizzi

PR #1669, opened 2023-05-24
[WINDUP-3881]-"Export zip" is not working in Windows

carlosthe19916:
https://issues.redhat.com/browse/WINDUP-3881

In Windows, the value of `String zipOutputPath` is formatted using `\` as a file separator; however, a URI object should be constructed by a String that uses `/` as a separator. Creating the final URI object based on a `java.io.File` makes sure the separators are handled automatically on each OS.


mrizzi (approved):
Changes have been tested successfully (thanks @PhilipCattanach).

Thanks for fixing this @carlosthe19916

PR #1670, opened 2023-05-24
[WINDUP-3883] - "Rules" page not loading in new PF4 reports [Windows]

carlosthe19916:
https://issues.redhat.com/browse/WINDUP-3883

- A minor number of Rules have Ids like `FileMapping_.*\.js$_FO`.
- Each rule is written in the disk in a single file where the filename is the ruleId
- Since rules like `FileMapping_.*\.js$_FO` contain invalid characters for Windows OS then those rules were not written in the disk generating an error in the UI-Rules Page


mrizzi (approved):
Changes have been tested successfully (thanks @PhilipCattanach).

Thanks for fixing this @carlosthe19916

PR #1673, opened 2023-05-31
Filter overriding rulesets like any other ruleset

mrizzi:


PhilipCattanach:
Hi @mrizzi, I retested with the latest commit and the override is still getting applied even if the override ruleset's target does  not match the execution target.

PhilipCattanach:
Tested successfully. Great new functionality, thanks @mrizzi

PR #1686, opened 2023-06-19
WINDUP-3928 Overriding rulesets always executed when in scope for the analysis

mrizzi:
https://issues.redhat.com/browse/WINDUP-3928


PhilipCattanach:
Tested successfully, well done

PR #1687, opened 2023-06-26
[WINDUP-3937] Add ability to decompile and analyse Java 17 apps

jmle:
See [Jira ticket](https://issues.redhat.com/browse/WINDUP-3937).

See the [ASM versions document](https://asm.ow2.io/versions.html) for more info on the dependency bump.


carlosthe19916 (approved):
I compared an application compiled in jdk11 and jdk17 and both generated the same data in the reports

PR #1713, opened 2023-12-21
WINDUP-4114 fix for HtmlUnit CVE

m-brophy:
https://issues.redhat.com/browse/WINDUP-4114
https://bugzilla.redhat.com/show_bug.cgi?id=2252942
https://access.redhat.com/security/cve/CVE-2023-49093

as agreed with @mrizzi I've opened this PR so we can decide if we want to make these changes to fix this CVE

Selenium versions updated in test-util/pom.xml as `v14.6.0` brings the correct htmlunit version, `v3.9.0`.
However this requires code changes because of htmlunit migration from v2->3, where pacvkage names change.
Even with this change, running tests brought a failure of `java.lang.IllegalArgumentException: Unknown HttpClient factory netty` wihen using test-util classes.
The code was using netty as html client only as a workaround for a bug in `jboss-modules:1.9.1.Final` affecting selenium, used by the latest version of furnace, https://issues.redhat.com/browse/MODULES-392.
So I overrode the version of jboss-modules to the fixed version, `v1,9,2,Final` which alowed us to use the selenium recommended http client,` jdk-http-client`

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -411,11 +411,11 @@
                 <artifactId>org.eclipse.core.resources</artifactId>
                 <version>[3.14.0,3.19.0)</version>
             </dependency>
-            <dependency>
+            <!--dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-all</artifactId>
                 <version>4.1.86.Final</version>
-            </dependency>
+            </dependency-->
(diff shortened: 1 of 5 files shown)


m-brophy:
test dependencies only so we've elected not to make these changes

PR #1739, opened 2024-05-10
WINDUP-4115 CVE upgrade css-tools

m-brophy:
https://issues.redhat.com/browse/WINDUP-4115
https://issues.redhat.com/browse/WINDUP-4098

https://bugzilla.redhat.com/show_bug.cgi?id=2254559

### DIFF
--- a/pf-ui/src/main/webapp/package.json
+++ b/pf-ui/src/main/webapp/package.json
@@ -3,7 +3,6 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@adobe/css-tools": "^4.3.3",
     "@migtools/lib-ui": "8.6.4",
     "@patternfly/patternfly": "4.224.2",
     "@patternfly/react-charts": "6.94.18",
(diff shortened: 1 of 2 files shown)


pf-ui/src/main/webapp/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@adobe/css-tools": "^4.3.3",
```
carlosthe19916, line 6: As @adobe/css-tools is not a direct dependency used by the UI it is better to locate this line under the "overrides" section of the package.json file

carlosthe19916 (changes requested):

pf-ui/src/main/webapp/package.json
m-brophy: Done!

carlosthe19916 (approved):
LGTM! Thanks @m-brophy

